PR #9, opened 2015-11-30
Update tables.sql.erb

kerkinos:
Add the project_account_history table


jimdowling:
Can you change the enum to a VARCHAR(255)?
We are following the pattern for enums: varchar in DB, Define an Enum in Java, use '@EnumAsString' in JPA. For an example, see the "ConsentsBean.java".

jimdowling:
Add default or NOT NULL qualifiers to each of the columns in the tables. You have no indexes on any of the tables. Define indexes on actions and date in both tables, and username in the first table.

PR #66, opened 2017-06-14
HOPSWORKS-74 change hostid in the hosts table from ip to hostname

berthoug:
https://hopshadoop.atlassian.net/browse/HOPSWORKS-74


jimdowling:
You're changing from the kagent::default IP to the hops::nn hostname. I assume you wanted the kagent::default hostnames?

PR #69, opened 2017-06-28
HOPSWORKS-97

kouzant:
Fix string interpolation in Chef template

https://hopshadoop.atlassian.net/browse/HOPSWORKS-97


kouzant:
There are, #{node['jupyter']['base_dir']} was not expanded to '/srv/hops/jupyter'

PR #130, opened 2017-12-20
HOPSWORKS-94

o-alex:
recipies for dela/hopssite. initial step to move out all dela specific things from hopsworks default to their specific recipies


tkakantousis:
@o-alex can you test this by following these instructions? http://hops.readthedocs.io/en/latest/developer_guide/guide/platform_testing.html

o-alex:
https://github.com/hopshadoop/hops-testing/pull/9

PR #161, opened 2018-03-16
HOPSWORKS-476 add zip support for datasets

robzor92:


robzor92:
https://github.com/hopshadoop/hops-testing/pull/63

PR #174, opened 2018-05-16
[HOPSWORKS-555] Fixes for upgrading Hopsworks

jimdowling:
https://hopshadoop.atlassian.net/browse/HOPSWORKS-555


Berksfile
```diff
@@ -4,31 +4,31 @@ source 'https://supermarket.chef.io'
 metadata
 
 cookbook 'java'
-cookbook 'conda', github: "hopshadoop/conda-chef", branch: "master"
-cookbook 'kagent', github: "hopshadoop/kagent-chef", branch: "master"
-cookbook 'hops', github: "hopshadoop/hops-hadoop-chef", branch: "master"
-cookbook 'ndb', github: "hopshadoop/ndb-chef", branch: "master"
+cookbook 'conda', github: "hopshadoop/conda-chef", branch: "jimfix"
```
tkakantousis: @jimdowling Reset branch to `master` before merging.

templates/default/systemd.service
```diff
@@ -16,6 +16,8 @@ Type = forking
 Environment="<%= key %>=<%= value %>"
 <% end -%>
 <% end -%>
+Restart=always
+TimeoutSec=420
```
SirOibaf: Remove the TimeoutSec from here. The correct place to change it is from the install recipe (which you did).

PR #179, opened 2018-06-04
[HOPSWORKS-578]  custom tensorflow binaries

jimdowling:
https://hopshadoop.atlassian.net/browse/HOPSWORKS-578

### DIFF
--- a/metadata.rb
+++ b/metadata.rb
@@ -728,6 +728,17 @@ attribute "tensorflow/git_url",
 :description => "url for git sourcecode for tensorflow",
 :type => "string"
 
+attribute "tensorflow/custom_url",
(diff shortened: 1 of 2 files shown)


metadata.rb
```diff
@@ -728,6 +728,17 @@
 :description => "url for git sourcecode for tensorflow",
 :type => "string"
 
+attribute "tensorflow/custom_url",
+          :description => "User-supplied URL for the tensorflow .whl binaries to be installed.",
+          :type => "string"
+
```
SirOibaf: Since Karamel 0.5, this is not required anymore. Having the attribute in tensorflow-chef is enough.

PR #182, opened 2018-07-19
[Hopsworks-621]

o-alex:
Expose the java recipe attributes to be used in cluster-definitions so we can set the exact java build we want to use.

### DIFF
--- a/metadata.rb
+++ b/metadata.rb
@@ -63,6 +63,16 @@ attribute "java/install_flavor",
           :display_name =>  "Oracle (default) or openjdk",
           :type => 'string'
 
+attribute "java/jdk/8/x86_64/url",
+          :display_name =>  "Oracle (default) or openjdk",
+          :type => 'string'
+attribute "java/jdk/8/x86_64/checksum",
+          :display_name =>  "Oracle (default) or openjdk",
+          :type => 'string'
+attribute "java/oracle/accept_oracle_download_terms",
+          :display_name =>  "Oracle (default) or openjdk",
+          :type => 'string'
+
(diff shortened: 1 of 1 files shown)


o-alex:
@jimdowling 
I tried setting in the hopsworks-chef de attributes/default:
`
default['java']['jdk_version'] = '8'
default['java']['install_flavor'] = 'oracle'
default['java']['jdk']['8']['x86_64']['url'] = 'http://download.oracle.com/otn-pub/java/jdk/8u181-b13/96a7b8442fe848ef90c96a2fad6ed6d1/jdk-8u181-linux-x64.tar.gz'
default['java']['jdk']['8']['x86_64']['checksum'] = '1845567095bfbfebd42ed0d09397939796d05456290fb20a83c476ba09f991d3'
default['java']['oracle']['accept_oracle_download_terms'] = true
`
Setting the above, did not make the other recipies see these default values. For example the hopsworks-monitor recipe would still see the default values set by the java recipe. Because of this, in order to get the fix working, I had to expose the java-recipe attributes and then add them to any cluster-definition that we want to use.

jimdowling:
Ok. The place to set them is not in hopsworks/attributes/default.rb.
It is in kagent/attributes/default.rb.
If you look at all of the attributes/default.rb files in all cookbooks, they all have:
include_attribute "kagent"
so they will inherit any default attribute values set in the kagent cookbook.
You do not need to set the metadata.rb in hopsworks-chef/metadata.rb, unless you want to change those values in a karamel cluster defn file.

o-alex:
no longer required - Hopsworks-621 only requires karamel-chef and kagent-chef PRs

PR #192, opened 2018-08-28
[HOPSWORKS-602] Update Hopsworks settings in cluster definition breaks flyway migrate

SirOibaf:


SirOibaf:
Testing: https://github.com/logicalclocks/hops-testing/pull/143

SirOibaf:
@jimdowling Yes, that's probably better. I updated the PR.

SirOibaf:
Actually no, we can't have a repeatable migration the reason is the following:
- version: 0.x introduces a new variable in the table
- version: 0.x+1 deletes that variable from the table 

if during an update 0.x gets modified (e.g. different ip) but 0.x+1 is not, then the variable will be added back in the settings as Flyway will execute 0.x but not 0.x+1.

I'm putting back the bash script to run all the dml scripts at every update.

PR #197, opened 2018-09-11
HOPSWORKS-660 fix zip/unzip setting wrong owner for files in same folder

robzor92:


robzor92:
https://github.com/logicalclocks/hops-testing/pull/153

templates/default/unzip-hdfs-files.sh.erb
```diff
@@ -59,7 +59,7 @@ echo "UNZIPPING" > "$fsmFile"
 
 
 cd "$workdir"
-<%= @dtrx %> -o -n "$file"
+dtrx -o -n "$file"
 check_error
```
SirOibaf: @robzor92 This was templated by @jimdowling in its last PR. Not completely sure why. But if you think that the template is not necessary, you should also remove the part in install recipe.

robzor92: Aha, the problem was that this is not valid syntax. It resolved to empty string in the script which broke everything. @jimdowling what's the intention behind this templating?

jimdowling: There is a reason for the templating. dtrx is not on the $PATH in centos - we install it from source, and it ends up in /usr/local/bin/dtrx
In ubuntu, it's installed as a package (probably in /usr/bin/dtrx).

jimdowling: From hopsworks::install.rb, you will see:
'dtrx' should not be an empty string.


install.rb:  package "dtrx"
install.rb:  dtrx="dtrx"
install.rb:  remote_file "#{Chef::Config['file_cache_path']}/dtrx.tar.gz" do
install.rb:    source node['download_url'] + "/#{node['dtrx']['version']}"
install.rb:  bash "unpack_dtrx" do
install.rb:    tar -xzf dtrx.tar.gz
install.rb:    cd dtrx-7.1
install.rb:    # dtrx expects 7z to on its path. create a symbolic link from /bin/7z to /bin/7za
install.rb:    not_if "which dtrx"
install.rb:  dtrx="/usr/local/bin/dtrx"
install.rb:              :dtrx => dtrx

recipes/install.rb
```diff
@@ -184,7 +185,7 @@
   package "dtrx"
   package "libkrb5-dev"
   dtrx="dtrx"
-when "redhat"
+when "rhel"
```
jimdowling: This was the bug!

recipes/install.rb
```diff
@@ -696,8 +697,10 @@
   owner node['glassfish']['user']
   group node['glassfish']['group']
   mode "550"
-  variables({
-              :dtrx => dtrx
+  variables(lazy {
```
jimdowling: This is not needed - the previous code was ok.

SirOibaf: No the previous code was not ok. If it was, we would see the empty string only on centos, but we see it empty also on ubuntu. The reason si that the value of the variable `dtrx` is only known at runtime, while the variables block is (without the lazy) evaluated at the beginning (when dtrx is still emtpy). See: https://docs.chef.io/resource_common.html#lazy-evaluation and

jimdowling: ok, sloppy by me, due to sloppy variable scoping in ruby/chef. dtrx should be defined before the platform case statement. The 'variables{lazy{' is not the simplest solution.

This should be the solution - first define the dtrx variable in the recipe scope, then override it in the rhel/centos case statement:

dtrx="dtrx"
switch platform_family:
case "debian":
# do nothing here
case "rhel":
dtrx="/usr/local/bin/dtrx"

PR #202, opened 2018-10-06
[HOPSWORKS-695] Dela update to hopsworks-ca v2 and elastic 6.2.3

o-alex:


o-alex:
jenkings testing https://github.com/logicalclocks/hops-testing/pull/175

templates/default/hopssite/hs_env.sh.erb
```diff
@@ -18,7 +18,7 @@ export DB_NAME='hops_site'
 export MYSQL_DIR='/srv/hops/mysql-cluster/ndb/scripts/'
 export MSQL_CONNECTOR='mysql-connector-java-5.1.29-bin.jar'
 export HOPS_SITE_WAR="${HOPS_SITE_BASE}/lib/hops-site.war"
-export HOPS_SITE_DOWNLOAD_URL="http://snurran.sics.se/hops/hops-site.war"
+export HOPS_SITE_DOWNLOAD_URL="http://snurran.sics.se/hops/hopssite/hops-site.war"
```
SirOibaf: Here you shouldn't hardcode snurran.sics.se. Use `node['download_url']` (or something like that) instead.

templates/default/hopssite/hs_rows.sql.erb
```diff
@@ -1,6 +1,6 @@
 insert into dataset_status select 0, "upload";
 insert into dataset_status select 1, "download";
-insert into variables select "dela_version", "0.1";
+insert into variables select "dela_version", "0.1.0";
```
SirOibaf: When and how is this script executed? Is it re-executed during an update? If so, you shouldn't use `insert into` as it will fail with a PK violation. Replace it with `REPLACE INTO`. See the rest of the dml files in /templates

templates/default/hopssite/hs_elastic.sh.erb
```diff
@@ -1,4 +1,4 @@
 #!/bin/bash
 set -e
 
-curl -XPUT '10.0.2.15:9200/hops-site?pretty' -H 'Content-Type: application/json' -d '{"mappings":{"dataset":{"dynamic":"strict","properties":{"description":{"type":"string"},"name":{"type":"string"}, "dsv":{"type":"integer"},"id":{"type":"string"},"xattr":{"type":"nested","dynamic":true}}}}}'
+curl -XPUT '10.0.2.15:9200/hops-site?pretty' -H 'Content-Type: application/json' -d '{"mappings":{"dataset":{"dynamic":"strict","properties":{"description":{"type":"text"},"name":{"type":"text"}, "dsv":{"type":"integer"},"id":{"type":"text"},"xattr":{"type":"nested","dynamic":true}}}}}'
```
SirOibaf: Why is this IP hardcoded into the script?

tkakantousis:
@o-alex HOPSWORKS-509 has been merged, so once @SirOibaf comments are addressed, we can merge it.

o-alex:
@tkakantousis @SirOibaf I addressed the issues above.
For the moment I think this is good enough. The next step is to take the hopssite(tracker) scripts out of hopsworks and into their own recipe. This will probably be the next PR.
I started a new jenkins test and once it succeeds this can be merged in.
https://github.com/logicalclocks/hops-testing/pull/175

PR #212, opened 2018-10-30
[HOPSWORKS-756] NPE when starting Jupyter

robzor92:


robzor92:
https://github.com/logicalclocks/hops-testing/pull/205

PR #220, opened 2018-11-08
[HOPSWORKS-793] TensorBoard too long-lived

robzor92:


robzor92:
https://github.com/logicalclocks/hops-testing/pull/220

PR #222, opened 2018-11-14
[HOPSWORKS-811] Conda speedup

jimdowling:
https://logicalclocks.atlassian.net/browse/HOPSWORKS-811

### DIFF
--- a/files/default/sql/ddl/0.6.1.sql
+++ b/files/default/sql/ddl/0.6.1.sql
@@ -1,1 +1,2 @@
 ALTER TABLE `hopsworks`.`project` ADD COLUMN `conda_env` TINYINT(1) DEFAULT 0;
+UPDATE `hopsworks`.`project` SET conda_env=1 WHERE conda=1;
--- a/files/default/sql/ddl/0.6.1.sql
+++ b/files/default/sql/ddl/0.6.1.sql
@@ -1,0 +1,1 @@
+ALTER TABLE `hopsworks`.`project` ADD COLUMN `conda_env` TINYINT(1) DEFAULT 0;
--- a/templates/default/sql/dml/0.6.1.sql.erb
+++ b/templates/default/sql/dml/0.6.1.sql.erb
@@ -1,1 +1,1 @@
-UPDATE `hopsworks`.`project` SET conda_env=1 WHERE conda=1;
+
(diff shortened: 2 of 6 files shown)


files/default/sql/ddl/0.6.1.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE `hopsworks`.`project` ADD COLUMN `conda_env` TINYINT(1) DEFAULT 0;
```
SirOibaf: Shouldn't this be `default 1`? On the current installations each project has a conda_env already

templates/default/sql/dml/0.6.1.sql.erb
```diff
@@ -0,0 +1 @@
+UPDATE `hopsworks`.`project` SET conda_env=1 WHERE conda=1;
```
SirOibaf: This is not how it works. The dml files are re-run at every update.

jimdowling: ok, will fix. We misunderstood it - thought they were run once. The idea behind flyway is that you can run an update between versions, and do an undo to rollback.

How do we do that now?

SirOibaf: First of all, the community version of Flyway doesn't support rollbacks, so if you want to rollback you have to do it manually.
For ddl instructions it works as before, except that they are just plain files, not templates.
For dml instructions we have 1 file per version and we run them all each time. The reason is that in the dml files we add IPs, framework/libraries versions and bunch of other attributes - mainly to the settings table. These attributes need to be updated at each update but we don't want to replicate the SQL in each dml version file because otherwise the files will grow exponentially.

Example: Let's say you add a settings which is used only by 0.6.1. Then you add to the `dml/0.6.1.sql` a `REPLACE INTO ...`. In the undo file you add the delete, so if you have to rollback (manually as for the ddl) you leave the settings table clean. The undo is particularly important if you change the meaning of a given field between updates.

jimdowling: I can't set default to '1'. It will screw up the code, as the code will then have to set it back to '0' in loads of places. '0' is the correct default.
What i want to do is set it to '1' for existing projects that have conda enabled already.
So i will move my dml command to the ddl file - even though the command itself is not strictly ddl. It should only be executed once. There is no undo for the update command, as dropping the column means the  updated value doesn't matter.

SirOibaf: Ok, this works as well.

PR #223, opened 2018-11-14
pip fix

robzor92:


jimdowling:
i fixed the same thing for the python kernel enable shell script. I don't think this needs changes in chef code. You just send down "env" as either the projectName or 'python27' or 'python36' from hopsworks.

robzor92:
Do we have a helper method for that? AnacondaDir returns the absolute path.

PR #232, opened 2018-12-27
[HOPSWORKS-851] Convert .ipynb to .py uses wrong jupyter binary path

robzor92:


robzor92:
https://github.com/logicalclocks/hops-testing/pull/249

PR #233, opened 2019-01-02
[HOPSWORKS-719] Implement feature store microservice UI on Hopsworks

Limmen:
- Add SQL for Featurestore Tables
- Add Featurestore Hive Dependency
- Add tfconnector spark dependency


Limmen:
Tests: https://github.com/logicalclocks/hops-testing/pull/248

files/default/sql/ddl/updates/0.7.0.sql
```diff
@@ -26,4 +26,4 @@ ALTER TABLE `hopsworks`.`ops_log` MODIFY COLUMN `inode_id` BIGINT(20);
 ALTER TABLE `hopsworks`.`meta_log` MODIFY COLUMN `meta_pk2` BIGINT(20);
 ALTER TABLE `hopsworks`.`meta_log` MODIFY COLUMN `meta_pk3` BIGINT(20);
 
-ALTER TABLE `hopsworks`.`conda_commands` ADD COLUMN `install_jupyter` tinyint(1) NOT NULL DEFAULT '0';
\ No newline at end of file
+ALTER TABLE `hopsworks`.`conda_commands` ADD COLUMN `install_jupyter` tinyint(1) NOT NULL DEFAULT '0';
```
SirOibaf: @Limmen Please don't modify ddl files for older releases, otherwise they will be re-remplated by chef with a different checksums and will break flyway migrations

templates/default/sql/dml/0.8.0.sql.erb
```diff
@@ -0,0 +1,3 @@
+REPLACE INTO variables VALUES ("featurestore_default_quota", "<%= @featurestore_default_quota %>");
+REPLACE INTO variables VALUES ("featurestore_default_storage_format", "<%= @featurestore_default_storage_format %>");
+REPLACE INTO variables VALUES ("tf_spark_connector_version", "<%= node['hops']['tf_spark_connector_version'] %>");
```
SirOibaf: You are missing the undo for these.

templates/default/sql/dml/0.8.0.sql.erb
```diff
@@ -0,0 +1,3 @@
+REPLACE INTO variables VALUES ("featurestore_default_quota", "<%= @featurestore_default_quota %>");
+REPLACE INTO variables VALUES ("featurestore_default_storage_format", "<%= @featurestore_default_storage_format %>");
```
SirOibaf: Here you are just passing a node attribute, instead of using a template variable, just put the node attribute as you do in the next line. 
It makes `hopsworks::default` recipe shorter and easier to read.

files/default/sql/ddl/updates/0.7.0.sql
Limmen: @SirOibaf Yeah good catch, it was unintentional, must have added a space or something when moving featurestore DDL/DML from 0.7.0 to 0.8.0

templates/default/sql/dml/0.8.0.sql.erb
Limmen: Done

templates/default/sql/dml/0.8.0.sql.erb
Limmen: Sure, done

PR #236, opened 2019-01-10
[HOPSWORKS-853] Feature Store Tour

Limmen:
- Move chef attributes for spark dependencies from hops-hadoop to spark-chef

### DIFF
--- a/templates/default/sql/dml/0.4.0.sql.erb
+++ b/templates/default/sql/dml/0.4.0.sql.erb
@@ -1,5 +1,6 @@
 DELETE FROM `hopsworks`.`variables` WHERE `id`='vagrant_enabled';
 DELETE FROM `hopsworks`.`variables` WHERE `id`='rest_port';
 
-REPLACE INTO `hopsworks`.`variables` VALUES ("hopsexamples_version", "<%= node['hops']['hopsexamples_version'] %>");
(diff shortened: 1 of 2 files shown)


Limmen:
Tests: https://github.com/logicalclocks/hops-testing/pull/253

templates/default/sql/dml/0.4.0.sql.erb
```diff
@@ -1,5 +1,6 @@
 DELETE FROM `hopsworks`.`variables` WHERE `id`='vagrant_enabled';
 DELETE FROM `hopsworks`.`variables` WHERE `id`='rest_port';
 
-REPLACE INTO `hopsworks`.`variables` VALUES ("hopsexamples_version", "<%= node['hops']['hopsexamples_version'] %>");
+REPLACE INTO `hopsworks`.`variables` VALUES ("hopsexamples_version", "<%= node['hadoop_spark']['hopsexamples_version']
```
tkakantousis: @Limmen Why is this in `0.4.0.sql.erb` and not in `0.9.0.sql.erb`?

tkakantousis: @SirOibaf flyway will only run `0.9.0.sql`, it skips previous versions for dml files. Is that correct or am I missing something?

tkakantousis: @Limmen I had a chat with @SirOibaf and you are right, I had misunderstood how we parse these files since we changed it recently. Will merge.

PR #241, opened 2019-01-28
[HOPSWORKS-887]

tkakantousis:


templates/default/sql/dml/0.9.0.sql.erb
```diff
@@ -0,0 +1 @@
+REPLACE INTO variables VALUES ("download_allowed", "<%= @download_allowed %>");
```
SirOibaf: should be `node['hopsworks']['download_allowed']` not `@download_allowed`

SirOibaf: Already fixed them in my other PR. In the other case, they are ugly, but they work (values are passed in the variables map in the hopsworks::default recipe). In your case `@download_allowed` is going to be completely empty

PR #242, opened 2019-01-30
[HOPSWORKS-894] InferenceController implementations should use connection pooling

robzor92:


metadata.rb
```diff
@@ -1422,10 +1422,18 @@
...
 ##
 ##
-## Presto
+## TFServing
 ##
 ##
 
+attribute "tfserving/pool_size",
+          :description => "size of the connection pool for serving inference requests",
+          :type => 'numeric'
```
SirOibaf: The current version of Karamel only likes string values

robzor92:
@SirOibaf @tkakantousis any other feedback? Otherwise this should be merged.

PR #244, opened 2019-01-31
[HOPSWORKS-546] Multiple Jupyter notebooks opened for the same user cause PersistenceException

robzor92:


robzor92:
https://github.com/logicalclocks/hops-testing/pull/276

files/default/sql/ddl/0.9.0__initial_tables.sql
```diff
@@ -617,6 +617,7 @@ CREATE TABLE `jupyter_project` (
   `pid` bigint(20) NOT NULL,
   `project_id` int(11) NOT NULL,
   PRIMARY KEY (`port`),
+  UNIQUE KEY `unique_hdfs_user` (`hdfs_user_id`),
   KEY `hdfs_user_idx` (`hdfs_user_id`),
```
SirOibaf: We should condense the `unique key` and the `key` in the `hdfs_user_id`. There is no point having both

files/default/sql/ddl/updates/0.9.0.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE jupyter_project ADD CONSTRAINT unique_hdfs_user UNIQUE (hdfs_user_id);
```
SirOibaf: Same as above

files/default/sql/ddl/0.9.0__initial_tables.sql
robzor92: from what I understand key and unique key are different. key creates a DB index while unique key is a constraint.

robzor92: fixed

files/default/sql/ddl/updates/0.9.0.sql
robzor92: fixed

files/default/sql/ddl/0.9.0__initial_tables.sql
SirOibaf: You cannot assume the name of the constraint. You have to query the information schema to get the correct name by querying the `information_schema` db

files/default/sql/ddl/updates/0.9.0.sql
```diff
@@ -0,0 +1,7 @@
...
+
+ALTER TABLE jupyter_project DROP FOREIGN KEY `FK_103_525`;
```
SirOibaf: You cannot assume the name of the constraint. You have to query the information schema to get the correct name by querying the `information_schema` db

robzor92: Added the sql for querying, migration and rollback both work

files/default/sql/ddl/updates/0.9.0.sql
```diff
@@ -0,0 +1,13 @@
+ALTER TABLE jupyter_project ADD CONSTRAINT unique_hdfs_user UNIQUE (hdfs_user_id);
+
+-- Find the name of the FK
+SET @fk_name = (SELECT CONSTRAINT_NAME FROM INFORMATION_SCHEMA.KEY_COLUMN_USAGE WHERE TABLE_SCHEMA = "hopsworks" AND TABLE_NAME = "jupyter_project" AND REFERENCED_TABLE_NAME="hdfs_users");
```
tkakantousis: @robzor92 Just fyi, that can return multiple constraints. In the future might be safer to add referenced column name as well. We can merge it.

PR #245, opened 2019-01-31
[HOPSWORKS-902] tensorboard.sh script uses hopworks user as group in chmod

robzor92:


templates/default/tensorboard.sh.erb
```diff
@@ -14,6 +14,7 @@ help() {
 
 ANACONDA_USER=<%= node['conda']['user'] %>
 HOPSWORKS_USER=<%= node['hopsworks']['user'] %>
```
SirOibaf: Is this still needed?

robzor92: Yes

robzor92: Removed it

PR #247, opened 2019-02-05
[HOPSWORKS-908] Airflow as a service

kouzant:
https://logicalclocks.atlassian.net/browse/HOPSWORKS-908

Check JIRA for dependencies


recipes/install.rb
```diff
@@ -313,6 +314,26 @@
               }
             }
           },
+          'airflowPool' => {
```
SirOibaf: If I remember correctly, during an update this part is not executed.

templates/default/convert-ipython-notebook.sh.erb
```diff
@@ -54,6 +54,14 @@ if [ $? -ne 0 ] ; then
...
 fi    
 
+
+echo "
+from pyspark import SparkContext
+sc = SparkContext(\"yarn\", \"${base}\")
+
+" > /tmp/"${base}".py
+cat "${base}".py >> /tmp/"${base}".py
+
```
SirOibaf: According to @robzor92  this not correct. The PySpark kernel creates a `SparkSession` object named `spark` not a `SparkContext` named sc.

robzor92: I am not opposed the idea. But as it is right now it is wrong if we want to make it compatible with Jupyter. It is as Fabio said. We should not get the **SparkContext**, but the **SparkSession**. And it **must** be named **spark**. You get the SparkSession like this. https://github.com/logicalclocks/hops-util-py/blob/7df48d2ecc6153513f6aed097335a16fa49b4ca9/hops/util.py#L408

robzor92: I am also assuming this is not tested. So unless it is done so properly after the changes, please remove this code. I will reintroduce it in HOPSWORKS-716 in that case.

kouzant: I removed this part of code

recipes/install.rb
kouzant: Code is there but since I can't test it on time, it will be merged on another commit.

PR #260, opened 2019-02-26
[HOPSWORKS-716] Unify Spark Configuration for Jobs service and Jupyter

robzor92:


robzor92:
https://github.com/logicalclocks/hops-testing/pull/313

robzor92:
@tkakantousis @SirOibaf Please review.

Berksfile
```diff
@@ -7,7 +7,7 @@ cookbook 'conda', github: "logicalclocks/conda-chef", branch: "master"
 cookbook 'kagent', github: "logicalclocks/kagent-chef", branch: "master"
 cookbook 'hops', github: "logicalclocks/hops-hadoop-chef", branch: "master"
 cookbook 'ndb', github: "logicalclocks/ndb-chef", branch: "master"
-cookbook 'hadoop_spark', github: "logicalclocks/spark-chef", branch: "master"
+cookbook 'hadoop_spark', github: "robzor92/spark-chef", branch: "HOPSWORKS-716"
```
SirOibaf: Non master dependency

files/default/sql/ddl/updates/0.10.0.sql
```diff
@@ -4,4 +4,29 @@ DROP TABLE IF EXISTS  `hopsworks`.`projectgenericuser_certs`;
...
+ALTER TABLE `jupyter_settings` DROP COLUMN `archives`;
+ALTER TABLE `jupyter_settings` DROP COLUMN `jars`;
+ALTER TABLE `jupyter_settings` DROP COLUMN `files`;
+ALTER TABLE `jupyter_settings` DROP COLUMN `py_files`;
+ALTER TABLE `jupyter_settings` DROP COLUMN `spark_params`;
+ALTER TABLE `jupyter_settings` DROP COLUMN `fault_tolerant`;
+ALTER TABLE `jupyter_settings` DROP COLUMN `umask`;
+ALTER TABLE `jupyter_settings` DROP COLUMN `log_level`;
+ALTER TABLE `jupyter_settings` ADD COLUMN `base_dir` VARCHAR(255) DEFAULT '/Jupyter/';
+ALTER TABLE `jupyter_settings` ADD COLUMN `json_config` TEXT NOT NULL;
```
SirOibaf: mmm, now sure here. Shouldn't you set the default to be emtpy string to handle already existing `jupyter_settings` entries?
Does the code handle empty settings, or should we write a migration step in Expat?

Berksfile
robzor92: Fixed

files/default/sql/ddl/updates/0.10.0.sql
robzor92: The code will handle it by setting the correct json_config so no need for a default value here. Particulary since it needs to be JSON of the class SparkJobConfiguration.

robzor92: No need for expat at all.

SirOibaf: Yeah, my bad. I was worried that MySQL would have complained. I ran the query on a VM and it worked fine.

PR #265, opened 2019-03-12
[HOPSWORKS-994] TensorBoard in experiments service fails to start if tensorflow-gpu is in the project environment

robzor92:


robzor92:
https://github.com/logicalclocks/hops-testing/pull/328

PR #274, opened 2019-04-01
[HOPSWORKS-1018] sqlContext should be injected in converted .ipynb file

robzor92:


robzor92:
sqlContext should be injected in converted .ipynb file

robzor92:
https://github.com/logicalclocks/hops-testing/pull/344

PR #285, opened 2019-05-22
[HOPSWORKS-1089] Integration of sparkmagic with maggy

jimdowling:
https://logicalclocks.atlassian.net/browse/HOPSWORKS-1089


files/default/sql/ddl/updates/0.10.0.sql
```diff
@@ -85,6 +85,24 @@ CREATE TABLE IF NOT EXISTS `oauth_login_state` (
...
+  `port` int(11) NOT NULL,
+  `secret` varchar(128) COLLATE latin1_general_cs NOT NULL,
+  `created`    TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
+  KEY (`app_id`, `port`),
+  PRIMARY KEY (`id`)
+)
+  ENGINE = ndbcluster
+  DEFAULT CHARSET = latin1
+  COLLATE = latin1_general_cs;
+
```
SirOibaf: Add the `DROP TABLE` in the 0.10.0_undo.sql

files/default/sql/ddl/updates/0.10.0.sql
```diff
@@ -85,6 +85,24 @@ CREATE TABLE IF NOT EXISTS `oauth_login_state` (
...
+--  Maggy Driver Service Lookup Table in Hopsworks
+--
+CREATE TABLE IF NOT EXISTS `maggy_driver` (
+  `id` MEDIUMINT NOT NULL AUTO_INCREMENT,
+  `app_id` char(30) COLLATE latin1_general_cs NOT NULL,
+  `host_ip` varchar(128) COLLATE latin1_general_cs NOT NULL,
+  `port` int(11) NOT NULL,
+  `secret` varchar(128) COLLATE latin1_general_cs NOT NULL,
+  `created`    TIMESTAMP DEFAULT CURRENT_TIMESTAMP,
+  KEY (`app_id`, `port`),
```
SirOibaf: I think here we should add `UNIQUE app_id` to avoid potential issues on the `getSingleResult()` in Hopsworks

files/default/sql/ddl/updates/0.10.0.sql
jimdowling: done

files/default/sql/ddl/updates/0.10.0.sql
jimdowling: after our discussion, i don't agree. It is incorrect usage of the experiments API if a client calls it twice in the same notebook. We will throw an exception that says you don't call it again.

PR #290, opened 2019-05-27
[HOPSWORKS-1095] Add back Localhost support

jimdowling:
https://logicalclocks.atlassian.net/browse/HOPSWORKS-1095


Berksfile
```diff
@@ -15,13 +15,13 @@ cookbook 'tensorflow', github: "logicalclocks/tensorflow-chef", branch: "master"
...
-cookbook 'kkafka', github: "logicalclocks/kafka-cookbook", branch: "master"
+cookbook 'kkafka', github: "logicalclocks/kafka-cookbook", branch: "localhost"
 cookbook 'elastic', github: "logicalclocks/elasticsearch-chef", branch: "master"
 cookbook 'hopslog', github: "logicalclocks/hopslog-chef", branch: "master"
 cookbook 'hopsmonitor', github: "logicalclocks/hopsmonitor-chef", branch: "master"
 cookbook 'glassfish', github: "logicalclocks/chef-glassfish", branch: "master"
-cookbook 'hops_airflow', github: "logicalclocks/airflow-chef", branch: "master"
-cookbook 'hive2', github: "logicalclocks/hive-chef", branch: "master"
+cookbook 'hops_airflow', github: "logicalclocks/airflow-chef", branch: "localhost"
+cookbook 'hive2', github: "logicalclocks/hive-chef", branch: "localhost"
```
SirOibaf: Need master here.

templates/default/sql/dml/0.10.0.sql.erb
```diff
@@ -39,3 +39,4 @@ REPLACE INTO `hopsworks`.`variables` VALUES ("hive_scratchdir_cleaner_interval",
 
 REPLACE INTO `hopsworks`.`variables` VALUES ("admin_email", "<%= node['hopsworks']['admin']['email'] %>");
 
+REPLACE INTO `hopsworks`.`variables` VALUES ("localhost", "<%= node['install']['localhost'] %>");
```
SirOibaf: What's the purpose of this setting? There is no PR in Hopsworks that uses it. 
If it's not used, let's just get rid of it.

jimdowling: I wanted to do a URL rewrite in the YarnProxyServlet. If 'localhost' is set, rewrite the target URL for the spark ui to localhost. This would relieve the need to rewrite /etc/hosts to point the fqdn to localhost.

SirOibaf: This `REPLACE` needs now to be moved to `1.0.0`

PR #304, opened 2019-07-04
[HOPSWORKS-1114] add validation key updated date and type

ErmiasG:


ErmiasG:
logicalclocks/hops-testing#441

PR #318, opened 2019-08-20
HOPSWORKS-640

tkakantousis:


recipes/default.rb
```diff
@@ -916,7 +918,7 @@
 directory node['hopsworks']['staging_dir'] + "/private_dirs"  do
   owner node['jupyter']['user']
   group node['hopsworks']['group']
-  mode "0330"
+  mode "0370"
```
tkakantousis: @kouzant Beamjobserver filebeat runs as glassfish user and needs to access all secret dirs, hence we give read permission. Is that ok?

kouzant: As long as we never add `jupyter` user to `glassfish` group, it's ok

PR #332, opened 2019-09-23
[HOPSWORKS-1140] Online Feature Store

Limmen:


Limmen:
Tests: https://github.com/logicalclocks/hops-testing/pull/507

recipes/default.rb
```diff
@@ -1128,3 +1128,59 @@
     EOF
   end
 end
+
+
+if node['mysql']['tls'].eql? "true"
```
SirOibaf: This doesn't belong here. It should be in ndb-chef

recipes/install.rb
```diff
@@ -316,6 +324,26 @@
               }
             }
           },
+          'featureStorePool' => {
```
SirOibaf: This is ok. *However* during an update, this part of the code won't be executed. So you need to create the connection pool also in `recipes/default`

PR #344, opened 2019-10-31
[HOPSWORKS-1388] Change text to varchar

tkakantousis:


files/default/sql/ddl/updates/1.1.0.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE `hopsworks`.`jobs` CHANGE COLUMN `json_config` `json_config` VARCHAR(12500) COLLATE latin1_general_cs NOT NULL;
+ALTER TABLE `hopsworks`.`jupyter_settings` CHANGE COLUMN `json_config` `json_config` VARCHAR(12500) COLLATE latin1_general_cs NOT NULL;
```
SirOibaf: 500 chars more than the one in the initial tables

PR #347, opened 2019-11-07
[HOPSWORKS-1384] Kafka REST API v2 - Schemas

kai-chi:


files/default/sql/ddl/1.1.0__initial_tables.sql
```diff
@@ -1112,6 +1112,51 @@ CREATE TABLE `project_team` (
...
+-- table structure for table `schemas`
+--
+
+/*!40101 set @saved_cs_client     = @@character_set_client */;
+/*!40101 set character_set_client = utf8 */;
+CREATE TABLE `schemas` (
+  `id` int(11) NOT NULL AUTO_INCREMENT,
+  `schema` varchar(10000) COLLATE latin1_general_cs NOT NULL,
+  `project_id` int(11) NOT NULL,
+  PRIMARY KEY (`id`, `project_id`),
```
SirOibaf: Why is `project_id` part of the primary key? `id` is enough as it's an `AUTO_INCREMENT`.

files/default/sql/ddl/1.1.0__initial_tables.sql
```diff
@@ -1112,6 +1112,51 @@ CREATE TABLE `project_team` (
...
+--
+-- table structure for table `schemas`
+--
+
+/*!40101 set @saved_cs_client     = @@character_set_client */;
+/*!40101 set character_set_client = utf8 */;
+CREATE TABLE `schemas` (
+  `id` int(11) NOT NULL AUTO_INCREMENT,
+  `schema` varchar(10000) COLLATE latin1_general_cs NOT NULL,
+  `project_id` int(11) NOT NULL,
```
SirOibaf: Can you clarify why we need the `project_id` in this table?

files/default/sql/ddl/1.1.0__initial_tables.sql
```diff
@@ -1112,6 +1112,51 @@ CREATE TABLE `project_team` (
...
+  PRIMARY KEY (`id`, `project_id`),
+  CONSTRAINT `project_idx` FOREIGN KEY (`project_id`) REFERENCES `project` (`id`) ON DELETE CASCADE ON UPDATE NO ACTION
+) ENGINE=ndbcluster DEFAULT CHARSET=latin1 COLLATE=latin1_general_cs;
+--
+-- table structure for table `subjects`
+--
+
+/*!40101 set @saved_cs_client     = @@character_set_client */;
+/*!40101 set character_set_client = utf8 */;
+CREATE TABLE `subjects` (
```
SirOibaf: How is sharing of schema (or subjects?) implemented in Hopsworks? Maybe I'm too tired, but I don't see how the is this schema (or subject?) shared with project x - is stored in the database.

files/default/sql/ddl/1.1.0__initial_tables.sql
kai-chi: Avro schemas will now be part of projects rather than global as it is now. `project_id` is necessary because the same schema can be registered from different projects

files/default/sql/ddl/1.1.0__initial_tables.sql
kai-chi: good call. it's not implemented atm because up to now the schemas were global. one can share a kafka topic with another project and everyone can use it but the second user would not have access to the topic's schema.
@tkakantousis, can we put it in another jira or should it be implemented in this one?

files/default/sql/ddl/1.1.0__initial_tables.sql
```diff
@@ -1112,6 +1112,51 @@ CREATE TABLE `project_team` (
...
+  KEY `created_on_idx` (`created_on`),
+  CONSTRAINT `project_idx` FOREIGN KEY (`project_id`) REFERENCES `project` (`id`) ON DELETE CASCADE ON UPDATE NO ACTION,
+  CONSTRAINT `schema_id_idx` FOREIGN KEY (`schema_id`, `project_id`) REFERENCES `schemas` (`id`, `project_id`) ON DELETE NO ACTION ON UPDATE NO ACTION
+) ENGINE=ndbcluster DEFAULT CHARSET=latin1 COLLATE=latin1_general_cs;
+/*!40101 SET character_set_client = @saved_cs_client */;
+
+--
+-- Table structure for table `subjects_compatibility`
+--
+CREATE TABLE `subjects_compatibility` (
```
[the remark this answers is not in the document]
kai-chi: it is not a field in `subjects` because there can be multiple versions of a subject but a common compatibility for all of them.
it does not have a constraint to `subjects` because one row in `subjects_compatibility` can refer to multiple rows in `subjects`. A foreign key identifies uniquely a row in another table so it's not a good candidate.
`schemas` and `subjects_compatibility` don't have a constraint because they are not related to each other.

files/default/sql/ddl/updates/1.1.0.sql
```diff
@@ -93,3 +93,185 @@ CREATE TABLE `statistic_columns` (
...
+    KEY `project_id_idx` (`project_id`),
+    KEY `created_on_idx` (`created_on`),
+    CONSTRAINT `project_idx` FOREIGN KEY (`project_id`)
+        REFERENCES `project` (`id`)
+        ON DELETE CASCADE ON UPDATE NO ACTION,
+    CONSTRAINT `schema_id_idx` FOREIGN KEY (`schema_id`)
+        REFERENCES `schemas` (`id`)
+        ON DELETE NO ACTION ON UPDATE NO ACTION,
+    CONSTRAINT `subjects__constraint_key` UNIQUE (`subject`, `version`, `project_id`)
+)  ENGINE=NDBCLUSTER DEFAULT CHARSET=LATIN1 COLLATE = LATIN1_GENERAL_CS;
```
tkakantousis: @kai-chi just for consistency, let's keep the values lower cased as in the other tables ` ENGINE=ndbcluster DEFAULT CHARSET=latin1 COLLATE=latin1_general_cs;`

PR #348, opened 2019-11-07
[HOPSWORKS-1380][featurestore] Configure Feature Group statistics during creation

moritzmeister:
https://logicalclocks.atlassian.net/browse/HOPSWORKS-1380
https://bbc1.sics.se/jenkins/job/HopsWorks-testing/1981/

### DIFF
--- a/files/default/sql/ddl/1.1.0__initial_tables.sql
+++ b/files/default/sql/ddl/1.1.0__initial_tables.sql
@@ -309,6 +309,13 @@ CREATE TABLE `feature_group` (
   `feature_group_type` INT(11) NOT NULL DEFAULT '0',
   `on_demand_feature_group_id` INT(11) NULL,
   `cached_feature_group_id` INT(11) NULL,
+  `desc_stats_enabled` TINYINT(1) NOT NULL DEFAULT '1',
+  `feat_corr_enabled` TINYINT(1) NOT NULL DEFAULT '1',
+  `feat_hist_enabled` TINYINT(1) NOT NULL DEFAULT '1',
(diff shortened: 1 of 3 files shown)


files/default/sql/ddl/1.1.0__initial_tables.sql
```diff
@@ -323,6 +330,22 @@ CREATE TABLE `feature_group` (
...
+--
+-- Table structure for table `statistic_columns`
+--
+
+/*!40101 SET @saved_cs_client     = @@character_set_client */;
+/*!40101 SET character_set_client = utf8 */;
+CREATE TABLE `statistic_columns` (
+  `id` int(11) NOT NULL AUTO_INCREMENT,
+  `feature_group_id` int(11) DEFAULT NULL,
+  `name` varchar(500) COLLATE latin1_general_cs DEFAULT NULL,
```
SirOibaf: How did you come up with 500? 
Hive column name max length is 767 chars while for MySQL is 64. I suppose we don't enforce this limit anywhere. I the new API we should enforce it though.

moritzmeister: https://github.com/logicalclocks/hopsworks-chef/blob/5955801ac989d5ef29dfeb2d1e2be58cbc954d69/files/default/sql/ddl/1.1.0__initial_tables.sql#L356

I was looking at other tables where we save feature/column-names.

SirOibaf: Ok

moritzmeister:
rebased on master, to avoid conflicts

PR #350, opened 2019-11-12
[HOPSWORKS-1356] migrate args from jobs to executions

tkakantousis:


tkakantousis:
@robzor92 please review

PR #352, opened 2019-11-22
[HOPSWORKS-1434] Scale jvm -Xmx for Spark and Jupyter based on allocated memory

robzor92:


robzor92:
https://github.com/logicalclocks/hops-testing/pull/554

PR #355, opened 2019-12-09
[HOPSWORKS-1372] Upgrade the ELK stack to 7.2 OSS and use the OpenDis…

maismail:
…tro Security Plugin
https://logicalclocks.atlassian.net/browse/HOPSWORKS-1372
tests: https://github.com/logicalclocks/hops-testing/pull/565

### DIFF
--- a/recipes/default.rb
+++ b/recipes/default.rb
@@ -21,7 +21,7 @@ hopsworks_db = "hopsworks"
 realmname = "kthfsrealm"
 
 begin
-  elastic_ips = node['elastic']['default']['private_ips'].join(",")
+  elastic_ips = all_elastic_ips_str()
 rescue
   elastic_ips = ""
   Chef::Log.warn "could not find the elastic server ip for HopsWorks!"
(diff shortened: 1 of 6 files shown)


metadata.rb
```diff
@@ -1334,7 +1334,57 @@
           :description =>  "Sha-1 checksum for the elasticsearch .tar.gz file",
           :type => 'string'
 
+attribute "elastic/opendistro_security/admin/username",
```
SirOibaf: There is no need to redefine all these attributes here - Since version 0.4 Karamel also checks the `metadata.rb` of the dependent cookbooks

recipes/default.rb
```diff
@@ -119,7 +119,7 @@
 end
 
 begin
-  kibana_ip = private_recipe_ip("hopslog","default")
+  kibana_ip = public_recipe_ip("hopslog","default")
```
SirOibaf: What is the reason for this change? We use everywhere the `private_recipe_ip`.

metadata.rb
maismail: Ok, good to know. I just added it here since i found that we are doing this for other cookbooks.

recipes/default.rb
maismail: Since we are removing the kibana proxy servlet and instead exposing a secure kibana to users.

PR #356, opened 2019-12-12
[HOPSWORKS-1474] Add reserved project names in the database

kouzant:
https://logicalclocks.atlassian.net/browse/HOPSWORKS-1474


attributes/default.rb
```diff
@@ -125,6 +125,7 @@
 default['hopsworks']['hive_default_quota_mbs']           = "250000"
 default['hopsworks']['featurestore_default_quota_mbs']   = "250000"
 default['hopsworks']['max_num_proj_per_user']            = "10"
+default['hopsworks']['reserved_project_names']           = "python27,python36,python37,python38,python39,hops-system,hopsworks,information_schema,airflow,glassfish_timers,grafana,hops,metastore,mysql,ndbinfo,performance_schema,sqoop,sys"
```
tkakantousis: @kouzant @SirOibaf shouldn't we add the os users as well?

kouzant: Why would we blacklist OS users?

PR #358, opened 2019-12-12
[HOPSWORKS-1481] Add endpoints to retrieve feature groups and trainin…

SirOibaf:
…g datasets by name


files/default/sql/ddl/1.1.0__initial_tables.sql
```diff
@@ -319,6 +319,7 @@ CREATE TABLE `executions` (
 /*!40101 SET character_set_client = utf8 */;
 CREATE TABLE `feature_group` (
   `id` int(11) NOT NULL AUTO_INCREMENT,
+  `name` varchar(255) NOT NULL,
```
moritzmeister: Now that we have the other unification, why `255`? Shouldn't it be the maximum of any of the types so existing table names don't get truncated?

We are limited by the online feature store to `63` in NDB anyway?

PR #360, opened 2020-01-02
[HOPSWORKS-1283] Zip does not work when there is a space in the file path

robzor92:


robzor92:
https://github.com/logicalclocks/hops-testing/pull/586

PR #362, opened 2020-01-08
[HOPSWORKS-1379] Timers should be rescheduled when failed and handle exceptions

robzor92:


robzor92:
https://github.com/logicalclocks/hops-testing/pull/587

attributes/default.rb
```diff
@@ -63,6 +63,8 @@
 default['hopsworks']['http_logs']['enabled']     = "true"
 default['hopsworks']['env_var_file']             = "#{node['hopsworks']['domains_dir']}/#{node['hopsworks']['domain_name']}_environment_variables"
 
+default['glassfish']['reschedule_failed_timer']     = "true"
```
tkakantousis: @robzor92 nitpick, shouldn't it be plural? `timers`

robzor92: I used the same naming scheme they do in the asadmin command, is it ok?

tkakantousis: of course, it seems more logical to me to have it in plural but lets stick to what they do :)

PR #363, opened 2020-01-10
[HOPSWORKS-1467][append] min provenance migrate/rollback (1.1.0)

o-alex:


o-alex:
I also added io.hops.hopsworks.expat.migrations.jobs.UpdateJobArgsConfiguration to expat since I think someone forgot to update the migration script. 
@tkakantousis is this ok, or should i not add the UpdateJobArgsConfiguration?

PR #366, opened 2020-01-14
[#34] variables for disabling password login and registration

ErmiasG:


SirOibaf:
Thanks, can you please also add the undo for the sql?

PR #371, opened 2020-01-24
[HOPSWORKS-1553] provenance cleaner timer fixes

o-alex:


o-alex:
@SirOibaf is the addition of a new variable considered database change?

PR #380, opened 2020-02-03
[HOPSWORKS-1382]-append add cron job to copy audit logs

ErmiasG:


recipes/default.rb
```diff
@@ -772,6 +772,19 @@
   end
 end
 
+if node['hopsworks']['audit_log_dump_enabled'].eql? "true"
```
kouzant: It's better if you use `casecmp` which makes case-insensitive comparison

templates/default/dump_audit_logs_to_hdfs.sh.erb
```diff
@@ -0,0 +1,79 @@
...
+
+LOGS_DIR=<%= @weblogs_dir %>
+HADOOP_HOME=<%= node['hops']['base_dir'] %>
+HDFS_WEBLOGS_DIR=<%= @remote_weblogs_dir %>
+
+CHECKPOINT=$LOGS_DIR/.checkpoint
+FILES_TO_DUMP=$LOGS_DIR/.files_to_dump
+REMOTE_FILE_PERMISSIONS=0750
+# This will match only files in the format of server_audit_log0.log
+REGEX=".*server_audit_log.*.log"
```
kouzant: How about switching the general 'any' character with only numbers?

ErmiasG: It can be with out any number too

recipes/default.rb
ErmiasG: do you mean 
if node['hopsworks']['audit_log_dump_enabled'].casecmp? "true"

ErmiasG: or
node['hopsworks']['audit_log_dump_enabled'].casecmp("true") == 0

PR #397, opened 2020-02-28
[HOPSWORKS-1405] Register Hopsworks with Consul and remove entries from variables table

kouzant:
https://logicalclocks.atlassian.net/browse/HOPSWORKS-1405

### DIFF
--- a/templates/default/sql/dml/1.3.0.sql.erb
+++ b/templates/default/sql/dml/1.3.0.sql.erb
@@ -1,3 +1,11 @@
 REPLACE INTO `hopsworks`.`tf_lib_mapping` VALUES ("1.15.0", "10.0", "7.6.0.64", "2.3");
 REPLACE INTO `hopsworks`.`variables` VALUES ("hdfs_base_storage_policy", "<%= node['hopsworks']['hdfs']['storage_policy']['base'] %>");
 REPLACE INTO `hopsworks`.`variables` VALUES ("hdfs_log_storage_policy", "<%= node['hopsworks']['hdfs']['storage_policy']['log'] %>");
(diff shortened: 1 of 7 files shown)


templates/default/sql/dml/1.3.0.sql.erb
```diff
@@ -1 +1,9 @@
-REPLACE INTO `hopsworks`.`tf_lib_mapping` VALUES ("1.15.0", "10.0", "7.6.0.64", "2.3");
\ No newline at end of file
+REPLACE INTO `hopsworks`.`tf_lib_mapping` VALUES ("1.15.0", "10.0", "7.6.0.64", "2.3");
+REPLACE INTO `hopsworks`.`variables` VALUES ("consul_domain", "<%= node['consul']['domain'] %>");
+DELETE FROM `hopsworks`.`variables` WHERE `id`='livy_ip';
```
SirOibaf: Since we remove all these IPs, can we also cleanup the `hopsworks::default` recipe from all these snippets:

```
begin
  rm_ip = private_recipe_ip("hops","rm")
rescue
  rm_ip = node['hostname']
  Chef::Log.warn "could not find the Resource Manager ip!"
end

begin
  rm_port = node['hops']['rm']['http_port']
rescue
  rm_port = 8088
  Chef::Log.warn "could not find the Resource Manager Port!"
end
```

kouzant: One think I'm not sure is whether it will fail templating previous DMLs which expect these variables

kouzant: I removed them

PR #415, opened 2020-04-06
[HOPSWORKS-1694] run users code in docker.

berthoug:
https://logicalclocks.atlassian.net/browse/HOPSWORKS-1694


berthoug:
See here for reviews already made before master-kube was removed: https://github.com/logicalclocks/hopsworks-chef/pull/421

PR #418, opened 2020-04-16
[HOPSWORKS-1678] Extend training dataset endpoint to save train-test split information as metadata

moritzmeister:
https://logicalclocks.atlassian.net/browse/HOPSWORKS-1678
https://bbc1.sics.se/jenkins/job/Hopsworks-testing-ee/189/


files/default/sql/ddl/1.3.0__initial_tables.sql
```diff
@@ -1532,6 +1532,7 @@ CREATE TABLE `training_dataset` (
   `hopsfs_training_dataset_id` INT(11) NULL,
   `external_training_dataset_id` INT(11) NULL,
   `training_dataset_type`   INT(11) NOT NULL DEFAULT '0',
+  `seed` INT(11) NULL,
```
SirOibaf, line 1535: As you set the Java type to `Long` this field should be `BigInt`

moritzmeister: fixed

PR #420, opened 2020-04-20
[HOPSWORKS-1657] System users x.509 v2

kouzant:
https://logicalclocks.atlassian.net/browse/HOPSWORKS-1657

### DIFF
--- a/recipes/install.rb
+++ b/recipes/install.rb
@@ -49,7 +49,7 @@ group node['hops']['hdfs']['user'] do
 end
 
 user node['hopsworks']['user'] do
-  home "/home/#{node['hopsworks']['user']}"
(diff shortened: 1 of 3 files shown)


recipes/install.rb
```diff
@@ -65,7 +65,7 @@
 end
 
 user node['hopsworks']['user'] do
-  home "/home/#{node['hopsworks']['user']}"
+  home node['hopsworks']['user-home']
```
ErmiasG, line 68: Is this defined somewhere? I only see ['glassfish']['user-home'] up there

kouzant: Interesting, I wonder how it worked. Anyway I changed it to `node['glassfish']['user-home']`

PR #421, opened 2020-04-21
[HOPSWORKS-1694] run users code in docker.

berthoug:
https://logicalclocks.atlassian.net/browse/HOPSWORKS-1694


files/default/sql/ddl/updates/1.3.0.sql
```diff
@@ -40,3 +40,10 @@ CREATE TABLE IF NOT EXISTS `training_dataset_split` (
   KEY `training_dataset_id` (`training_dataset_id`),
   CONSTRAINT `training_dataset_fk` FOREIGN KEY (`training_dataset_id`) REFERENCES `training_dataset` (`id`) ON DELETE CASCADE ON UPDATE NO ACTION
 ) ENGINE=ndbcluster DEFAULT CHARSET=latin1 COLLATE=latin1_general_cs;
+ALTER TABLE `hopsworks`.`python_dep` ADD COLUMN `base_env` VARCHAR(45) COLLATE latin1_general_cs;
+ALTER TABLE `hopsworks`.`conda_commands` DROP FOREIGN KEY `FK_481_519`;
```
SirOibaf, line 44: `FK_481_519` is a random id generated by MySQL if you create a constraint without name. So there might be deployments out there that have a different name for the same fk.

The correct way of dropping this is to look up the name in the `INFORMATION_SCHEMA` - here is an example: https://github.com/logicalclocks/hopsworks-chef/blob/d70685ba5aea4f77689e42b65ffad3650d27d92e/files/default/sql/ddl/updates/1.0.0.sql#L221-L226

files/default/sql/ddl/updates/undo/1.3.0__undo.sql
```diff
@@ -16,3 +14,12 @@ ALTER TABLE `hopsworks`.`host_services` CHANGE COLUMN `name` `service` varchar(4
...
 
 DROP TABLE IF EXISTS `training_dataset_split`;
+ALTER TABLE `hopsworks`.`python_dep` DROP COLUMN `base_env`;
+ALTER TABLE `hopsworks`.`conda_commands` ADD COLUMN `host_id` int(11) NOT NULL;
+ALTER TABLE `hopsworks`.`conda_commands` ADD KEY (`host_id`);
+ALTER TABLE `hopsworks`.`conda_commands` ADD CONSTRAINT `FK_481_519` FOREIGN KEY (`host_id`) REFERENCES `hosts` (`id`) ON DELETE CASCADE ON UPDATE NO ACTION ;
+ALTER TABLE `hopsworks`.`conda_commands` CHANGE `docker_image` `proj` varchar(255) COLLATE latin1_general_cs NOT NULL;
+ALTER TABLE `hopsworks`.`jupyter_project` CHANGE `cid` `pid` bigint(20) NOT NULL;
+ALTER TABLE `hopsworks`.`tensorboard` CHANGE `cid` `pid` bigint(20) NOT NULL;
+ALTER TABLE `hopsworks`.`tensorboard` CHANGE `cid` `local_pid` bigint(20) NOT NULL;
```
SirOibaf, line 24: this should be `serving`

metadata.rb
```diff
@@ -1430,14 +1430,6 @@
 ##
 ##
 
-attribute "jupyter/user",
-          :description => "Jupyter username to run notebook servers as",
-          :type => 'string'
-
-attribute "jupyter/group",
```
SirOibaf, line 1437: Could you also remove the `serving/user` and `serving/group`?

recipes/default.rb
```diff
@@ -307,7 +315,8 @@
          :dela_ip => dela_ip,
          :nonconda_hosts_list => nonconda_hosts_list,
          :krb_ldap_auth => node['ldap']['enabled'].to_s == "true" || node['kerberos']['enabled'].to_s == "true",
-         :featurestore_jdbc_url => featurestore_jdbc_url
+         :featurestore_jdbc_url => featurestore_jdbc_url,
+         :registry_host => registry_host
```
SirOibaf, line 319: Do we need to store this in the variables at all? Can't we have Hopsworks itself looking it up with the consul client?

recipes/install.rb
```diff
@@ -686,9 +666,12 @@
   action :create
 end
 
-["convert-ipython-notebook.sh", "anaconda-rsync.sh", "zip-hdfs-files.sh", "zip-background.sh",
-  "unzip-background.sh", "anaconda-command-ssh.sh", "conda-command-ssh.sh", "tensorboard.sh", "tensorboard-launch.sh", 
-  "tensorboard-cleanup.sh", "condasearch.sh", "pipsearch.sh", "list_environment.sh"].each do |script|
+["anaconda-rsync.sh", "zip-hdfs-files.sh", "zip-background.sh",
+ "unzip-background.sh", "anaconda-command-ssh.sh", "conda-command-ssh.sh",
+ "tensorboard-launch.sh", "tensorboard-cleanup.sh", "condasearch.sh",
```
SirOibaf, line 671: I don't think we need all these scripts anymore - like `anaconda-rsync`, `anaconda-command-ssh`, `conda-command-ssh`, "list_environment.sh", "jupyter-kill.sh", "jupyter-launch.sh",
 "tfserving-kill.sh", "sklearn_serving-launch.sh", "sklearn_serving-kill.sh"

templates/default/dockerImage.sh.erb
```diff
@@ -0,0 +1,60 @@
...
+
+elif [ "$1" == "delete" ] ; then
+    # Sanity checks for injection attacks
+    re='^([a-z0-9]+(-[a-z0-9]+)*\.)*[a-z0-9]+(:[0-9]*)?\/[_a-zA-Z0-9]{0,62}[a-zA-Z0-9]$'
+    if ! [[ $2 =~ $re ]] ; then
+	echo "error: Not an image name $3" >&2
+	help
+    fi
+
+    docker rmi $2
```
SirOibaf, line 36: Can we remove the image from the registry as well?

templates/default/sql/dml/1.3.0.sql.erb
```diff
@@ -11,3 +11,10 @@ DELETE FROM `hopsworks`.`variables` WHERE `id`='hiveserver_ext_hostname';
 DELETE FROM `hopsworks`.`variables` WHERE `id`='hiveserver_ssl_hostname';
 REPLACE INTO `hopsworks`.`variables` VALUES ("hopsworks_enterprise", "<%= node['install']['enterprise']['install']%>");
 REPLACE INTO `hopsworks`.`variables` VALUES ("kubernetes_installed", "<%= node['install']['kubernetes'] %>");
+INSERT INTO `hopsworks`.`variables` VALUES ("yarn_runtime", "docker");
+INSERT INTO `hopsworks`.`variables` VALUES ("docker_mounts", "<%= node['hops']['docker']['mounts'] %>");
+REPLACE INTO `hopsworks`.`variables` VALUES ("jupyter_user", "<%= node['hops']['yarnapp']['user'] %>");
+REPLACE INTO `hopsworks`.`variables` VALUES ("jupyter_group", "<%= node['hops']['group'] %>");
+REPLACE INTO variables VALUES ("anaconda_dir", "/srv/hops/anaconda");
```
SirOibaf, line 18: You should build the path based on the node attributes

templates/default/sql/dml/1.3.0.sql.erb
```diff
@@ -11,3 +11,10 @@ DELETE FROM `hopsworks`.`variables` WHERE `id`='hiveserver_ext_hostname';
 DELETE FROM `hopsworks`.`variables` WHERE `id`='hiveserver_ssl_hostname';
 REPLACE INTO `hopsworks`.`variables` VALUES ("hopsworks_enterprise", "<%= node['install']['enterprise']['install']%>");
 REPLACE INTO `hopsworks`.`variables` VALUES ("kubernetes_installed", "<%= node['install']['kubernetes'] %>");
+INSERT INTO `hopsworks`.`variables` VALUES ("yarn_runtime", "docker");
+INSERT INTO `hopsworks`.`variables` VALUES ("docker_mounts", "<%= node['hops']['docker']['mounts'] %>");
```
SirOibaf, line 15: Please use `REPLACE INTO` as these files are run at every upgrade. So the next upgrade there will be a PK violation if you just use `INSERT INTO`

templates/default/sql/dml/1.3.0.sql.erb
```diff
@@ -11,3 +11,10 @@ DELETE FROM `hopsworks`.`variables` WHERE `id`='hiveserver_ext_hostname';
...
+REPLACE INTO `hopsworks`.`variables` VALUES ("jupyter_user", "<%= node['hops']['yarnapp']['user'] %>");
+REPLACE INTO `hopsworks`.`variables` VALUES ("jupyter_group", "<%= node['hops']['group'] %>");
+REPLACE INTO variables VALUES ("anaconda_dir", "/srv/hops/anaconda");
+INSERT INTO `hopsworks`.`variables` VALUES ("docker_registry_name", "<%= @registry_host %>");
+INSERT INTO `hopsworks`.`variables` VALUES ("docker_registry_port", "<%= node['hops']['docker']['registry']['port'] %>");
```
SirOibaf, line 20: These should be discoverable from consul as well (cc. @kouzant )

templates/default/sql/dml/undo/1.3.0__undo.sql.erb
```diff
@@ -4,3 +4,9 @@ DELETE FROM `hopsworks`.`variables` WHERE `id`='hdfs_log_storage_policy';
 DELETE FROM `hopsworks`.`variables` WHERE `id`='service_discovery_domain';
 DELETE FROM `hopsworks`.`variables` WHERE `id`='hopsworks_enterprise';
 DELETE FROM `hopsworks`.`variables` WHERE `id`='kubernetes_installed';
+DELETE FROM `hopsworks`.`variables` WHERE `id`='yarn_runtime';
+REPLACE INTO variables VALUES ("anaconda_dir", "<%= node['conda']['base_dir'] %>");
+DELETE FROM `hopsworks`.`variables` WHERE `id`='docker_mounts';
+REPLACE INTO variables VALUES ("anaconda_dir", "<%= node['conda']['base_dir'] %>");
```
SirOibaf, line 10: double replace for `anacond_dir`, also please be consistent and use `hopsworks`.`variables`.

templates/default/sql/dml/1.3.0.sql.erb
SirOibaf: can you use the full name `hopswork`.`variables` here as well?

files/default/sql/ddl/updates/1.3.0.sql
berthoug: I took the FK_481_519 from here: https://github.com/logicalclocks/hopsworks-chef/blob/1c401873137f1cb8d05e29a86f7c988ba6466abc/files/default/sql/ddl/1.3.0__initial_tables.sql#L180
So it seems to come from us. But I'll do what you say, to be on the safe side.

files/default/sql/ddl/updates/undo/1.3.0__undo.sql
berthoug: thanks. Fixing it.

metadata.rb
berthoug: doing that. I don't know why I missed it.

recipes/default.rb
berthoug: Consul did not exist when I started on this work. Can we do it in a future Jira?

recipes/install.rb
berthoug: I think a bunch of these scripts is used by kagent. As I have not touched kagent in this work (cleaning it up will be a future jira) I have preferred to let the removal of these scripts to the kagent work.

templates/default/sql/dml/undo/1.3.0__undo.sql.erb
berthoug: Fixing it. 
It is also broken by the new collumn that have been added to the variables table.

templates/default/dockerImage.sh.erb
berthoug: It's on my todo list. It needs to be done through a rest call so it will probably be done in hopsworks directly but the final design is not decided yet.

templates/default/sql/dml/1.3.0.sql.erb
berthoug: fixed

templates/default/sql/dml/1.3.0.sql.erb
berthoug: fixed

templates/default/sql/dml/1.3.0.sql.erb
berthoug: Consul did not exist when I started this work. Can we please get this in and then improve it in future jiras with the help of the people that know the code?

PR #423, opened 2020-04-28
[HOPSWORKS-1721] add visibility level to variables

SirOibaf:


templates/default/sql/dml/undo/0.2.0__undo.sql.erb
```diff
@@ -1,2 +0,0 @@
-drop table `jim`;
```
tkakantousis, line 1: :D

PR #434, opened 2020-05-11
[HOPSWORKS-1314] It should be possible to start Jupyter with a Shared Dataset as base directory

robzor92:


tkakantousis:
Let's merge this after HOPSWORKS-1691

PR #436, opened 2020-05-13
HOPSWORKS-1779 add kube master

tkakantousis:


tkakantousis:
> please also add the include in the `attributes/default.rb` and the `depends` in the `metradata.rb`

thanks, done

PR #446, opened 2020-05-27
[HOPSWORKS-1805] consul for docker registry

berthoug:
https://logicalclocks.atlassian.net/browse/HOPSWORKS-1805

### DIFF
--- a/recipes/default.rb
+++ b/recipes/default.rb
@@ -246,14 +246,6 @@ unless node['install']['cloud'].strip.empty?
   node.override['hopsworks']['reserved_project_names'] = "#{node['hopsworks']['reserved_project_names']},cloud"
 end
 
-begin
-  registry_ip = private_recipe_ip("hops","docker_registry")
-  registry_host = resolve_hostname(registry_ip)
-rescue
-  registry_host = "localhost"
-  Chef::Log.warn "could not find the docker registry ip!"
-end
-
 for version in versions do
   # Template DML files
   template "#{theDomain}/flyway/dml/V#{version}__hopsworks.sql" do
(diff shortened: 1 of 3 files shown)


recipes/default.rb
```diff
@@ -286,8 +278,6 @@
          :dela_ip => dela_ip,
          :krb_ldap_auth => node['ldap']['enabled'].to_s == "true" || node['kerberos']['enabled'].to_s == "true",
          :featurestore_jdbc_url => featurestore_jdbc_url,
-         :registry_host => registry_host,
-         :hops_version => get_hops_version(node['hops']['version'])
```
tkakantousis, line 290: @berthoug why is the hops_version removed?

templates/default/sql/dml/undo/1.4.0__undo.sql.erb
```diff
@@ -1,8 +1,6 @@
 DELETE FROM `hopsworks`.`variables` WHERE `id`='yarn_runtime';
 REPLACE INTO `hopsworks`.`variables` (`id`, `value`) VALUES ("anaconda_dir", "<%= node['conda']['base_dir'] %>");
 DELETE FROM `hopsworks`.`variables` WHERE `id`='docker_mounts';
-DELETE FROM `hopsworks`.`variables` WHERE `id`='docker_registry_name';
-DELETE FROM `hopsworks`.`variables` WHERE `id`='docker_registry_port';
```
tkakantousis, line 5: Shouldn't the undo put back the values?
```
REPLACE INTO `hopsworks`.`variables` (`id`, `value`) VALUES ("docker_registry_name", "<%= @registry_host %>");	
REPLACE INTO `hopsworks`.`variables` (`id`, `value`) VALUES ("docker_registry_port", "<%= node['hops']['docker']['registry']['port'] %>");
```

berthoug: No, we are removing the fact that the values are being added to the database completely (adding the value has been done in the same "release" as we are now) so it will be as if these values had never existed.

recipes/default.rb
berthoug: probably a mistake when fixing conflicts. I'm fixing it.

PR #475, opened 2020-08-12
[HOPSWORKS-1945] Store AWS access/secret key in the secret store.

gibchikafa:


files/default/sql/ddl/updates/1.4.0.sql
```diff
@@ -67,4 +67,11 @@ ALTER TABLE `hopsworks`.`dataset_shared_with` ADD COLUMN `permission` VARCHAR(45
...
 ALTER TABLE `hopsworks`.`activity` CHANGE COLUMN `activity` `activity` VARCHAR(255) COLLATE latin1_general_cs NOT NULL;
 
-UPDATE `hopsworks`.`anaconda_repo` SET `url`="pypi" WHERE `url`="PyPi";
\ No newline at end of file
+UPDATE `hopsworks`.`anaconda_repo` SET `url`="pypi" WHERE `url`="PyPi";
+
+ALTER TABLE `hopsworks`.`feature_store_s3_connector` DROP COLUMN `access_key`;
+ALTER TABLE `hopsworks`.`feature_store_s3_connector` DROP COLUMN `secret_key`;
+
+ALTER TABLE `hopsworks`.`feature_store_s3_connector` MODIFY `name` VARCHAR(150) COLLATE latin1_general_cs  NOT NULL;
```
moritzmeister, line 75: @SirOibaf is this something we should be doing? It's probably unlikely, but what if someone has a connector with a longer name in his deployment and wants to upgrade hopsworks?

moritzmeister (commented):
It would be good if @SirOibaf can also review this quickly.

files/default/sql/ddl/updates/1.4.0.sql
SirOibaf: I don't think anybody has a storage connector named more than 150 chars, so I think it's fine.

SirOibaf:
I'm ok with merging this PR, @gibchikafa  we need to move the database changes to the 2.0.0 files. It's *extremely important* that all the changes (including empty lines ad so on) are removed from the 1.4.0 files.

SirOibaf (approved):

PR #476, opened 2020-08-14
[HOPSWORKS-1580] Online Feature Store JDBC password retrieval from UI

gibchikafa:


SirOibaf:
@gibchikafa - does this PR actually adds anything which is not added in the other PRs?

PR #478, opened 2020-08-17
HOPSWORKS-1806 delete tags from docker registry

tkakantousis:


templates/default/dockerImage.sh.erb
```diff
@@ -27,18 +27,13 @@ if [ "$1" == "create" ] ; then
     docker push $3
 
 elif [ "$1" == "delete" ] ; then
-    # Sanity checks for injection attacks
-    if ! [[ $2 =~ $VALID_IMAGE_NAME ]] ; then
-	echo "error: Not an image name $3" >&2
-	help
-    fi
+    # Remove tags from docker registry
+    docker exec registry rm -rf /var/lib/registry/docker/registry/v2/repositories/$2/_manifests/tags/$3
```
maismail, line 31: Can you only add these calls to the script only if managed_docker_registry is false

templates/default/dockerImage.sh.erb
```diff
@@ -27,18 +27,13 @@ if [ "$1" == "create" ] ; then
...
+
+elif [ "$1" == "gc" ] ; then
```
maismail, line 33: Same here

templates/default/dockerImage.sh.erb
tkakantousis: I'd prefer to keep the script as is and not call it for `delete` and `gc`
https://github.com/logicalclocks/hopsworks-ee/pull/179/files#diff-856e3ba78886504a2386eefbb49a8261R145

tkakantousis: We need another registryGC implementation for the managed one

maismail: Even though, we will need to add the bash code here to do a delete in case of ECR for example 

`aws ecr batch-delete-image --repository-name my-repo --image-ids imageTag=latest
`

maismail (changes requested):

maismail (approved):

robzor92 (approved):

PR #481, opened 2020-08-26
[HOPSWORKS-1975] Add support for passing job arguments in the Airflow workflow builder

gibchikafa:


files/default/hopsworks_templates/airflow_dag.py
```diff
@@ -60,6 +60,7 @@
 					 project_name="${operator.projectName}",
 					 task_id="${operator.id}",
 					 job_name="${operator.jobName}",
+                     job_arguments="${operator.jobArgs}",
```
moritzmeister, line 63: Nitpick, can you adjust the indentation to match the other arguments?

moritzmeister (changes requested):

files/default/hopsworks_templates/airflow_dag.py
gibchikafa: Fixed

moritzmeister (approved):

PR #489, opened 2020-09-03
[HOPSWORKS-2001] TFServing should use base image instead of the hops one

robzor92:


robzor92:
https://github.com/logicalclocks/hops-testing-ee/pull/162

tkakantousis (approved):

PR #491, opened 2020-09-09
[HOPSWORKS-2013] Cleanup docker images from managed registries

maismail:
https://logicalclocks.atlassian.net/browse/HOPSWORKS-2013
tests: https://github.com/logicalclocks/hops-testing-ee/pull/168

### DIFF
--- a/templates/default/dockerImage.sh.erb
+++ b/templates/default/dockerImage.sh.erb
@@ -26,20 +26,15 @@ if [ "$1" == "create" ] ; then
 
     docker push $3
 
-elif [ "$1" == "delete" ] ; then
 <% if node['install']['managed_docker_registry'] == "false" -%>
+elif [ "$1" == "delete" ] ; then
     # Remove tags from docker registry
     docker exec registry rm -rf /var/lib/registry/docker/registry/v2/repositories/$2/_manifests/tags/$3
-<% else -%>
-    # Not supported yet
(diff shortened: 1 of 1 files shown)


templates/default/dockerImage.sh.erb
```diff
@@ -26,20 +26,15 @@ if [ "$1" == "create" ] ; then
...
     docker exec registry rm -rf /var/lib/registry/docker/registry/v2/repositories/$2/_manifests/tags/$3
-<% else -%>
-    # Not supported yet
-<% end -%>
 
 elif [ "$1" == "gc" ] ; then
-<% if node['install']['managed_docker_registry'] == "false" -%>
     # Run garbage collection for registry
     docker exec registry bin/registry garbage-collect /etc/docker/registry/config.yml -m
-<% else -%>
```
tkakantousis, line 41: So this script is not going to be used for non-managed ? If so, why not removing this? `<% if node['install']['managed_docker_registry'] == "false" -%>`

tkakantousis (commented):
So

templates/default/dockerImage.sh.erb
maismail: To avoid adding non usable code to the dockerImage.sh in case of managed docker installations

tkakantousis (approved):

PR #501, opened 2020-09-24
[HOPSWORKS-2025] Delete job executions

tkakantousis:
We need this for integration tests in ruby

### DIFF
--- a/templates/default/sql/dml/2.0.0.sql.erb
+++ b/templates/default/sql/dml/2.0.0.sql.erb
@@ -1,0 +1,1 @@
+REPLACE INTO `hopsworks`.`variables`(`id`, `value`) VALUES ("rmyarn_user", "<%= node['hops']['rm']['user'] %>");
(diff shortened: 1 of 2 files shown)


templates/default/sql/dml/2.0.0.sql.erb
```diff
@@ -0,0 +1 @@
+REPLACE INTO `hopsworks`.`variables`(`id`, `value`) VALUES ("rmyarn_user", "<%= node['hops']['rm']['user'] %>");
```
robzor92, line 1: Missing the undo.

robzor92 (changes requested):

templates/default/sql/dml/2.0.0.sql.erb
tkakantousis: fixed

robzor92 (approved):

PR #505, opened 2020-10-01
[HOPSWORKS-2057] Drop condacommands-project fk

tkakantousis:
When a project is deleted, there need to be "REMOVE" commands left in the table but with the FK constraint these commands would be deleted by the db. Hopsworks ensures that any stray commands will be eventually removed (details in the Hopsworks PR of this JIRA)


SirOibaf:
Closing this PR - Moved the changes to  #507

PR #515, opened 2020-10-09
[HOPSWORKS-1847] Use MySQL JDBC to setup online storage connectors

SirOibaf:


recipes/default.rb
```diff
@@ -617,7 +607,7 @@
   end
 end
 
-glassfish_asadmin "create-jdbc-connection-pool --restype javax.sql.DataSource --datasourceclassname com.mysql.jdbc.jdbc2.optional.MysqlDataSource --ping=true --isconnectvalidatereq=true --validationmethod=auto-commit --description=\"Featurestore connection pool\" --property user=#{node['featurestore']['user']}:password=#{node['featurestore']['password']}:url=#{featurestore_jdbc_url_escaped} featureStorePool" do
+glassfish_asadmin "create-jdbc-connection-pool --restype javax.sql.DataSource --datasourceclassname com.mysql.jdbc.jdbc2.optional.MysqlDataSource --ping=true --isconnectvalidatereq=true --validationmethod=auto-commit --description=\"Featurestore connection pool\" --property user=#{node['featurestore']['user']}:password=#{node['featurestore']['password']}:url=\"#{node['featurestore']['jdbc_url'].sub!(":", "\\:")}\" featureStorePool" do
```
kouzant, line 610: Previously, we were substituting `:` with `\\\\:` Doesn't it need to be escaped here too?

kouzant (approved):

recipes/default.rb
SirOibaf: I tried with `irb` and it works correctly. Except that I should have used `gsub` instead of `sub` as sub substitutes only the first occurrence.

kouzant (approved):

PR #516, opened 2020-10-09
[HOPSWORKS-2014]

ErmiasG:


ErmiasG:
logicalclocks/hops-testing-ee/pull/197

attributes/default.rb
```diff
@@ -442,4 +442,5 @@
 default['hopsworks']['enable_metadata_designer']              = "false"
 
 #check nodemanager status
-default["hopsworks"]['check_nodemanager_status']              = "false"
\ No newline at end of file
+default["hopsworks"]['check_nodemanager_status']              = "false"
+default['hopsworks']['cloud']['type']                         = "NONE"
```
tkakantousis, line 446: What's the difference with https://github.com/logicalclocks/hopsworks-chef/blob/master/templates/default/sql/dml/1.0.0.sql.erb#L3 ? Seems they are both used to store the cloud type (aws/gcs/azure)

tkakantousis (changes requested):

attributes/default.rb
ErmiasG: Are you sure it is used to store aws/gcs/azure?

tkakantousis (approved):

PR #519, opened 2020-10-14
[HOPSWORKS-2083] be off dr-elephant

jimdowling:
https://logicalclocks.atlassian.net/browse/HOPSWORKS-2083


SirOibaf (changes requested):
More cleanup is required here, otherwise deployment will fail. We can get rid of the code in `hopsworks::default` and in the dml files.

SirOibaf (approved):

PR #522, opened 2020-10-26
[HOPSWORKS-2044] Add option to return an error message if a user starts a YARN job but…

gibchikafa:
… no nodemanagers are available

### DIFF
--- a/attributes/default.rb
+++ b/attributes/default.rb
@@ -440,3 +440,6 @@ default['hopsworks']['hdfs']['storage_policy']['base']        = "DB"
 default['hopsworks']['hdfs']['storage_policy']['log']         = "HOT"
 
 default['hopsworks']['enable_metadata_designer']              = "false"
+
+#check nodemanager status
+default["hopsworks"]['check_nodemanager_status']              = "false"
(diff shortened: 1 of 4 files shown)


templates/default/sql/dml/2.0.0.sql.erb
```diff
@@ -1,3 +1,5 @@
 REPLACE INTO `hopsworks`.`variables`(`id`, `value`) VALUES ("rmyarn_user", "<%= node['hops']['rm']['user'] %>");
 
 REPLACE INTO `hopsworks`.`variables` (`id`, `value`) VALUES ("spark_executor_min_memory", 1024);
+
+REPLACE INTO `hopsworks`.`variables` (`id`, `value`) VALUES ("check_nodemanagers_status", false);
```
tkakantousis, line 5: the value should come from an variable defined in attributes/default.rb and also defined in metadata.rb. Default value is "false"

tkakantousis (changes requested):

attributes/default.rb
```diff
@@ -439,4 +439,7 @@
 # To not fill the SSDs with Logs files that nobody access frequently we set the StoragePolicy for the LOGS dir to be default HOT
 default['hopsworks']['hdfs']['storage_policy']['log']         = "HOT"
 
-default['hopsworks']['enable_metadata_designer']              = "false"
\ No newline at end of file
+default['hopsworks']['enable_metadata_designer']              = "false"
+
+#check nodemanager status
+default["hopsworks"]['check_nodemanager_status']              = "false"
```
tkakantousis, line 445: Needs to be added in metadata.rb as well so that it can be overwritten in the cluster definition

tkakantousis (changes requested):

tkakantousis (approved):

PR #532, opened 2020-11-20
[HOPSWORKS-2163] Add missing sql changes

tkakantousis:


files/default/sql/ddl/updates/2.1.0.sql
```diff
@@ -0,0 +1,5 @@
+SET @fk_name = (SELECT CONSTRAINT_NAME FROM INFORMATION_SCHEMA.KEY_COLUMN_USAGE WHERE TABLE_SCHEMA = "hopsworks" AND TABLE_NAME = "conda_commands" AND REFERENCED_TABLE_NAME="project");
+SET @s := concat('ALTER TABLE hopsworks.conda_commands DROP FOREIGN KEY `', @fk_name, '`');
```
SirOibaf, line 2: Does this query run if the foreign key does not exists?

tkakantousis: It will fail with `ERROR 1064 (42000): You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'NULL' at line 1`.

Why would the foreign key not exist?

SirOibaf: In case someone is updating from 1.4.1

tkakantousis: fixed

SirOibaf (approved):

PR #537, opened 2020-11-26
[HOPSWORKS-2131] Add pip check command to Python service

robzor92:


files/default/sql/ddl/2.1.0__initial_tables.sql
```diff
@@ -1014,21 +1014,39 @@ CREATE TABLE `project` (
...
   `docker_image` varchar(255) COLLATE latin1_general_cs DEFAULT NULL,
+  `python_env_id` int(11) DEFAULT NULL,
   PRIMARY KEY (`id`),
   UNIQUE KEY `projectname` (`projectname`),
   UNIQUE KEY `inode_pid` (`inode_pid`,`inode_name`,`partition_id`),
   KEY `user_idx` (`username`),
   CONSTRAINT `FK_262_290` FOREIGN KEY (`username`) REFERENCES `users` (`email`) ON DELETE NO ACTION ON UPDATE NO ACTION,
-  CONSTRAINT `FK_149_289` FOREIGN KEY (`inode_pid`,`inode_name`,`partition_id`) REFERENCES `hops`.`hdfs_inodes` (`parent_id`,`name`,`partition_id`) ON DELETE CASCADE ON UPDATE NO ACTION
+  CONSTRAINT `FK_149_289` FOREIGN KEY (`inode_pid`,`inode_name`,`partition_id`) REFERENCES `hops`.`hdfs_inodes` (`parent_id`,`name`,`partition_id`) ON DELETE CASCADE ON UPDATE NO ACTION,
+  CONSTRAINT `FK_362_309` FOREIGN KEY (`python_env_id`) REFERENCES `python_environment` (`id`) ON DELETE SET NULL ON UPDATE NO ACTION
```
tkakantousis, line 1031: let's make the constraint names more human-friendly

files/default/sql/ddl/2.1.0__initial_tables.sql
```diff
@@ -1014,21 +1014,39 @@ CREATE TABLE `project` (
...
+--
+
+/*!40101 SET @saved_cs_client     = @@character_set_client */;
+/*!40101 SET character_set_client = utf8 */;
+CREATE TABLE `python_environment` (
+  `id` int(11) NOT NULL AUTO_INCREMENT,
+  `project_id` int(11) NOT NULL,
+  `python_version` varchar(25) COLLATE latin1_general_cs NOT NULL,
+  `jupyter_conflicts` TINYINT(1) NOT NULL DEFAULT '0',
+  `conflicts` VARCHAR(10000) COLLATE latin1_general_cs DEFAULT NULL,
```
tkakantousis, line 1046: we can go up to 12-13000?

files/default/sql/ddl/2.1.0__initial_tables.sql
```diff
@@ -1014,21 +1014,39 @@ CREATE TABLE `project` (
...
+--
+-- Table structure for table `python_environment`
+--
+
+/*!40101 SET @saved_cs_client     = @@character_set_client */;
+/*!40101 SET character_set_client = utf8 */;
+CREATE TABLE `python_environment` (
+  `id` int(11) NOT NULL AUTO_INCREMENT,
+  `project_id` int(11) NOT NULL,
+  `python_version` varchar(25) COLLATE latin1_general_cs NOT NULL,
```
tkakantousis, line 1044:
```suggestion
  `python_version` VARCHAR(25) COLLATE latin1_general_cs NOT NULL,
```

files/default/sql/ddl/2.1.0__initial_tables.sql
tkakantousis, line 1031: Since we now have a new table for keeping track for python envs for projects, having a ` UNIQUE KEY `project_env` (`project_id`,`python_version`),` means we can have multiple environments per project. How does this FK help in maintaining constraint integrity? It seems it forces a project to have one environment or am I missing something? The child table is `python_environment`, it is an index scan to find the python environment of a project which is good enough performance wise I believe

tkakantousis (changes requested):

files/default/sql/ddl/2.1.0__initial_tables.sql
robzor92: done

files/default/sql/ddl/2.1.0__initial_tables.sql
robzor92: went to 12k

files/default/sql/ddl/2.1.0__initial_tables.sql
robzor92: done

files/default/sql/ddl/2.1.0__initial_tables.sql
robzor92: I changed the UNIQUE KEY to only be used on the project_id now. As we don't have the backend implementation for multiple environments I don't see a direct need to model the database for that right now. The old key was a leftover.

tkakantousis: Still, when will the `project.python_env_id` be used? 
The project's env is derived from the dimension table `python_environment` and when a project is removed, the environment will be removed due to the FK. 

In JPA there owning side is PythonEnviromnent and in Project there will be the mapped entity (PythonEnvironment).

Similar to 
```
@OneToMany(cascade = CascadeType.ALL,
      mappedBy = "projectId")
  private Collection<JupyterProject> jupyterProjectCollection;
```
but with OneToOne.

robzor92: You're right there is not a need to have a foreign key in this case. I got rid of it.

robzor92:
https://github.com/logicalclocks/hops-testing-ee/pull/245

files/default/sql/ddl/2.1.0__initial_tables.sql
```diff
@@ -1032,6 +1031,24 @@ CREATE TABLE `project` (
...
+/*!40101 SET character_set_client = utf8 */;
+CREATE TABLE `python_environment` (
+  `id` int(11) NOT NULL AUTO_INCREMENT,
+  `project_id` int(11) NOT NULL,
+  `python_version` VARCHAR(25) COLLATE latin1_general_cs NOT NULL,
+  `jupyter_conflicts` TINYINT(1) NOT NULL DEFAULT '0',
+  `conflicts` VARCHAR(12000) COLLATE latin1_general_cs DEFAULT NULL,
+  UNIQUE KEY `project_env` (`project_id`),
+  PRIMARY KEY (`id`),
+  CONSTRAINT `FK_PROJECT_ID` FOREIGN KEY (`project_id`) REFERENCES `project` (`id`) ON DELETE CASCADE ON UPDATE NO ACTION
```
tkakantousis, line 1048:
```suggestion
  CONSTRAINT `FK_PYTHONENV_PROJECT` FOREIGN KEY (`project_id`) REFERENCES `project` (`id`) ON DELETE CASCADE ON UPDATE NO ACTION
```
just to make it a bit more unique

files/default/sql/ddl/2.1.0__initial_tables.sql
```diff
@@ -1032,6 +1031,24 @@ CREATE TABLE `project` (
...
+) ENGINE=ndbcluster AUTO_INCREMENT=119 DEFAULT CHARSET=latin1 COLLATE=latin1_general_cs;
```
tkakantousis, line 1049: why `AUTO_INCREMENT=119`?

tkakantousis (changes requested):

files/default/sql/ddl/2.1.0__initial_tables.sql
robzor92: fixed

files/default/sql/ddl/2.1.0__initial_tables.sql
robzor92: removed it, don't need to set it

tkakantousis (approved):

PR #543, opened 2020-12-17
[HOPSWORKS-2177] On demand feature groups over S3 connector

SirOibaf:


files/default/sql/ddl/updates/undo/2.1.0__undo.sql
```diff
@@ -105,4 +105,6 @@ SET SQL_SAFE_UPDATES = 1;
 ALTER TABLE `hopsfs_training_dataset` DROP FOREIGN KEY `hopsfs_td_conn_fk`,
     DROP COLUMN `connector_id`;
 
-DROP TABLE `feature_store_connector`;
\ No newline at end of file
+ALTER TABLE `on_demand_feature_group` DROP COLUMN `data_format`, DROP COLUMN `path`;
```
moritzmeister, line 108: What about the ```MODIFY `query` VARCHAR(11000), ``` shouldn't that one also be undone? e.g. set the NOT NULL constraint again.

SirOibaf: done

moritzmeister (approved):

PR #548, opened 2021-01-12
[HOPSWORKS-CLOUD-138] Upgrade prepartions

maismail:


recipes/default.rb
```diff
@@ -261,6 +261,21 @@
     mode 0750
     action :create
   end
+
+   # template all the ddl files from all versions
+   cookbook_file "#{theDomain}/flyway/all/sql/V#{version}__hopsworks.sql" do
+    source "sql/ddl/updates/#{version}.sql"
+    owner node['glassfish']['user']
```
kouzant, line 268: Explicitly set group of file

recipes/default.rb
```diff
@@ -261,6 +261,21 @@
...
+    mode 0750
+    action :create
+  end
+
+  cookbook_file "#{theDomain}/flyway/all/undo/U#{version}__undo.sql" do
+    source "sql/ddl/updates/undo/#{version}__undo.sql"
+    owner node['glassfish']['user']
```
kouzant, line 275: Same here, set group of file

kouzant (changes requested):

recipes/default.rb
maismail: Ok, done. However, I was keeping it the same as other blocks in the recipe https://github.com/logicalclocks/hopsworks-chef/blob/master/recipes/default.rb#L205

kouzant: I'm not sure if the assigned group would be the primary group of the `owner` or the primary group of the user which runs karamel

kouzant (approved):

PR #549, opened 2021-01-12
[HOPSWORKS-2280] Save statistics configuration also for training datasets

moritzmeister:
There is one issue @SirOibaf the following columns should have been removed in the 2.0 release. Probably a bad refactor. they don't hurt in the table, however, I add drop column statements for them in the 2.1.0 update sql, which will fail if they exist. I am wondering if we should just leave them for 2.0 clusters or fix the release branch etc?

```
  `cluster_analysis_enabled` TINYINT(1) NOT NULL DEFAULT '1',
  `num_clusters` int(11) NOT NULL DEFAULT '5',
  `num_bins` INT(11) NOT NULL DEFAULT '20',
  `corr_method` VARCHAR(50) NOT NULL DEFAULT 'pearson',
```

### DIFF
--- a/files/default/sql/ddl/2.1.0__initial_tables.sql
+++ b/files/default/sql/ddl/2.1.0__initial_tables.sql
@@ -329,13 +329,6 @@ CREATE TABLE `feature_group` (
   `feature_group_type` INT(11) NOT NULL DEFAULT '0',
   `on_demand_feature_group_id` INT(11) NULL,
   `cached_feature_group_id` INT(11) NULL,
-  `desc_stats_enabled` TINYINT(1) NOT NULL DEFAULT '1',
-  `feat_corr_enabled` TINYINT(1) NOT NULL DEFAULT '1',
-  `feat_hist_enabled` TINYINT(1) NOT NULL DEFAULT '1',
-  `cluster_analysis_enabled` TINYINT(1) NOT NULL DEFAULT '1',
-  `num_clusters` int(11) NOT NULL DEFAULT '5',
-  `num_bins` INT(11) NOT NULL DEFAULT '20',
-  `corr_method` VARCHAR(50) NOT NULL DEFAULT 'pearson',
   PRIMARY KEY (`id`),
   UNIQUE KEY `name_version` (`feature_store_id`, `name`, `version`),
   KEY `feature_store_id` (`feature_store_id`),
@@ -350,6 +343,27 @@ CREATE TABLE `feature_group` (
 /*!40101 SET character_set_client = @saved_cs_client */;
 
 --
+-- Table structure for table `statistics_config`
+--
+
+/*!40101 SET @saved_cs_client     = @@character_set_client */;
+/*!40101 SET character_set_client = utf8 */;
+CREATE TABLE `statistics_config` (
+  `id` int(11) NOT NULL AUTO_INCREMENT,
+  `feature_group_id` int(11),
+  `training_dataset_id` int(11),
+  `descriptive` TINYINT(1) NOT NULL DEFAULT '1',
+  `correlations` TINYINT(1) NOT NULL DEFAULT '1',
+  `histograms` TINYINT(1) NOT NULL DEFAULT '1',
+  PRIMARY KEY (`id`),
+  KEY `feature_group_id` (`feature_group_id`),
+  KEY `training_dataset_id` (`training_dataset_id`),
+  CONSTRAINT `fg_statistics_config_fk` FOREIGN KEY (`feature_group_id`) REFERENCES `feature_group` (`id`) ON DELETE CASCADE ON UPDATE NO ACTION,
+  CONSTRAINT `td_statistics_config_fk` FOREIGN KEY (`training_dataset_id`) REFERENCES `training_dataset` (`id`) ON DELETE CASCADE ON UPDATE NO ACTION
+) ENGINE=ndbcluster DEFAULT CHARSET=latin1 COLLATE=latin1_general_cs;
+/*!40101 SET character_set_client = @saved_cs_client */;
+
+--
 -- Table structure for table `statistic_columns`
 --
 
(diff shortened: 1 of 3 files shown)


SirOibaf:
> There is one issue @SirOibaf the following columns should have been removed in the 2.0 release. Probably a bad refactor. they don't hurt in the table, however, I add drop column statements for them in the 2.1.0 update sql, which will fail if they exist. I am wondering if we should just leave them for 2.0 clusters or fix the release branch etc?
> 
> ```
>   `cluster_analysis_enabled` TINYINT(1) NOT NULL DEFAULT '1',
>   `num_clusters` int(11) NOT NULL DEFAULT '5',
>   `num_bins` INT(11) NOT NULL DEFAULT '20',
>   `corr_method` VARCHAR(50) NOT NULL DEFAULT 'pearson',
> ```

No it's not safe to fix the release branch, but in the update file you can do `DROP COLUMN IF EXISTS` so that the migration doesn't fail if the column is missing already

files/default/sql/ddl/2.1.0__initial_tables.sql
```diff
@@ -1516,6 +1530,9 @@ CREATE TABLE `training_dataset` (
   `training_dataset_type`   INT(11) NOT NULL DEFAULT '0',
   `seed` BIGINT(11) NULL,
   `query` TINYINT(1) NOT NULL DEFAULT '0',
+  `desc_stats_enabled` TINYINT(1) NOT NULL DEFAULT '1',
+  `feat_corr_enabled` TINYINT(1) NOT NULL DEFAULT '1',
+  `feat_hist_enabled` TINYINT(1) NOT NULL DEFAULT '1',
```
SirOibaf, line 1535: Why are these here?

files/default/sql/ddl/updates/2.1.0.sql
```diff
@@ -147,4 +147,65 @@ CREATE TABLE `on_demand_option` (
...
+       `desc_stats_enabled` as `descriptive`,
+       `feat_corr_enabled` as `correlations`,
+       `feat_hist_enabled` as `histograms`
+    FROM `hopsworks`.`feature_group`;
+
+ALTER TABLE `hopsworks`.`feature_group`
+    DROP COLUMN `cluster_analysis_enabled`,
+    DROP COLUMN `num_clusters`,
+    DROP COLUMN `num_bins`,
+    DROP COLUMN `corr_method`,
```
SirOibaf, line 177: `IF EXISTS`

files/default/sql/ddl/updates/2.1.0.sql
```diff
@@ -147,4 +147,65 @@ CREATE TABLE `on_demand_option` (
...
+  `name` varchar(500) COLLATE latin1_general_cs DEFAULT NULL,
+  PRIMARY KEY (`id`),
+  KEY `statistics_config_id` (`statistics_config_id`),
+  CONSTRAINT `statistics_config_fk` FOREIGN KEY (`statistics_config_id`) REFERENCES `statistics_config` (`id`) ON DELETE CASCADE ON UPDATE NO ACTION
+) ENGINE=ndbcluster DEFAULT CHARSET=latin1 COLLATE=latin1_general_cs;
+
+INSERT INTO `hopsworks`.`statistic_columns` (`statistics_config_id`, `name`)
+SELECT * FROM `hopsworks`.`tmp_stats_columns`;
+
+DROP TABLE `hopsworks`.`tmp_stats_columns`;
```
SirOibaf, line 210: This looks complicated. Here's what I'd do:
- add `statistics_config_id` to the `statistics_column` table.
- Update the rows so that `statistics_config_id` matches the correct value based on the `feature_group_id` already present
- Drop the `feature_group_id` column.

files/default/sql/ddl/updates/undo/2.1.0__undo.sql
```diff
@@ -112,3 +112,39 @@ ALTER TABLE `on_demand_feature_group`
...
+  `name` varchar(500) COLLATE latin1_general_cs DEFAULT NULL,
+  PRIMARY KEY (`id`),
+  KEY `feature_group_id` (`feature_group_id`),
+  CONSTRAINT `statistic_column_fk` FOREIGN KEY (`feature_group_id`) REFERENCES `feature_group` (`id`) ON DELETE CASCADE ON UPDATE NO ACTION
+) ENGINE=ndbcluster DEFAULT CHARSET=latin1 COLLATE=latin1_general_cs;
+
+INSERT INTO `hopsworks`.`statistic_columns` (`feature_group_id`, `name`)
+SELECT * FROM `hopsworks`.`tmp_stats_columns`;
+
+DROP TABLE `hopsworks`.`tmp_stats_columns`;
```
SirOibaf, line 150: Same comment as the above.

SirOibaf (changes requested):

SirOibaf:
> > There is one issue @SirOibaf the following columns should have been removed in the 2.0 release. Probably a bad refactor. they don't hurt in the table, however, I add drop column statements for them in the 2.1.0 update sql, which will fail if they exist. I am wondering if we should just leave them for 2.0 clusters or fix the release branch etc?
> > ```
> >   `cluster_analysis_enabled` TINYINT(1) NOT NULL DEFAULT '1',
> >   `num_clusters` int(11) NOT NULL DEFAULT '5',
> >   `num_bins` INT(11) NOT NULL DEFAULT '20',
> >   `corr_method` VARCHAR(50) NOT NULL DEFAULT 'pearson',
> > ```
> 
> No it's not safe to fix the release branch, but in the update file you can do `DROP COLUMN IF EXISTS` so that the migration doesn't fail if the column is missing already

Actually MySQL doesn't seem to support the `DROP COLUMN IF EXISTS` statement - https://chase-seibert.github.io/blog/2010/01/15/mysql-drop-column-if-exists.html - we need to query the information schema.

files/default/sql/ddl/updates/2.1.0.sql
moritzmeister: yeah mysql doesn't support that, will do it with querying the schema

files/default/sql/ddl/2.1.0__initial_tables.sql
moritzmeister: thanks, that's a leftover, removed them.

files/default/sql/ddl/updates/2.1.0.sql
moritzmeister: simplified it as suggested.

SirOibaf (approved):

PR #552, opened 2021-01-15
[HOPSWORKS-2131] Append sql migration to new python_environment table

robzor92:


files/default/sql/ddl/updates/2.1.0.sql
```diff
@@ -163,6 +163,18 @@ CREATE TABLE IF NOT EXISTS `python_environment` (
   CONSTRAINT `FK_PYTHONENV_PROJECT` FOREIGN KEY (`project_id`) REFERENCES `project` (`id`) ON DELETE CASCADE ON UPDATE NO ACTION
 ) ENGINE=ndbcluster DEFAULT CHARSET=latin1 COLLATE=latin1_general_cs;
 
+INSERT INTO `hopsworks`.`python_environment` (`project_id`, `python_version`)
+SELECT `id`, `python_version`
+FROM `hopsworks`.`project`
+WHERE `python_version` IS NOT NULL
+AND `conda` = true;
+
+SET SQL_SAFE_UPDATES = 0;
```
tkakantousis, line 172: I thought `SQL_SAFE_UPDATES` is used for update queries without a `where` clause. Did you try it without turning SQL_SAFE_UPDATES off?

tkakantousis (approved):

files/default/sql/ddl/updates/2.1.0.sql
robzor92: It doesn't work as the UPDATE itself does not have a WHERE clause. The WHERE clause you see there is in the inner SELECT so I need to disable the `SQL_SAFE_UPDATES` to run the UPDATE statement.

robzor92: I've run the migration and rollback and couple of times now with projects that have python enabled and disabled, and redeployed hopsworks to see if it works. This code is now good to merge.

tkakantousis: oh yes you are correct

PR #553, opened 2021-01-17
[HOPSWORKS-2197] Add ADLS storage connector for the feature store

SirOibaf:


files/default/sql/ddl/updates/2.1.0.sql
```diff
@@ -54,13 +67,15 @@ CREATE TABLE IF NOT EXISTS `feature_store_connector` (
   `s3_id`                   INT(11),
   `hopsfs_id`               INT(11),
   `redshift_id`             INT(11),
+  `adls_id`                 INT(11),
```
moritzmeister, line 70: I think you forgot the undo for this column.

moritzmeister (changes requested):

files/default/sql/ddl/updates/2.1.0.sql
SirOibaf: Actually we don't need the undo for that one, as the `feature_store_connector` table was introduced in 2.1.0, so in the UNDO we already have:
```
DROP TABLE `feature_store_connector`;
```

moritzmeister: aaah, sorry I missed that.

moritzmeister (approved):

PR #562, opened 2021-01-30
[HOPSWORKS-2305] Remove security questions, user address, organizatio…

SirOibaf:
…n and ssh keys

### DIFF
--- a/files/default/sql/ddl/2.2.0__initial_tables.sql
+++ b/files/default/sql/ddl/2.2.0__initial_tables.sql
@@ -399,16 +399,6 @@ CREATE TABLE `files_to_remove` (
 /*!40101 SET character_set_client = @saved_cs_client */;
 
 --
--- Temporary table structure for view `hops_users`
---
-
-SET @saved_cs_client     = @@character_set_client;
-SET character_set_client = utf8;
-/*!50001 CREATE VIEW `hops_users` AS SELECT
- 1 AS `project_user`*/;
-SET character_set_client = @saved_cs_client;
-
---
 -- Table structure for table `hopssite_cluster_certs`
 --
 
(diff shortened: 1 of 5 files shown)


files/default/sql/ddl/2.2.0__initial_tables.sql
```diff
@@ -1804,23 +1716,6 @@ CREATE TABLE `variables` (
...
-/*!50001 SET @saved_col_connection     = @@collation_connection */;
-/*!50001 SET character_set_client      = utf8 */;
-/*!50001 SET character_set_results     = utf8 */;
-/*!50001 SET collation_connection      = utf8_general_ci */;
-/*!50001 CREATE ALGORITHM=UNDEFINED */
-/*!50001 VIEW `hops_users` AS select concat(`pt`.`team_member`,'__',`p`.`projectname`) AS `project_user` from ((`project` `p` join `project_team` `pt`) join `ssh_keys` `sk`) where `pt`.`team_member` in (select `u`.`email` from (`users` `u` join `ssh_keys` `s`) where (`u`.`uid` = `s`.`uid`)) */;
-/*!50001 SET character_set_client      = @saved_cs_client */;
-/*!50001 SET character_set_results     = @saved_cs_results */;
-/*!50001 SET collation_connection      = @saved_col_connection */;
-
```
kouzant, line 1823: How about these? Don't we need the view

SirOibaf: I have no idea where this `hops_users` was used. I can't find it anywhere. It's also weird that the view is referring to the `ssh_keys` table. I had to remove it since I drop that table.

kouzant (approved):

PR #570, opened 2021-02-17
garbage collection for acr

lovew-lc:
Allows deletion of images inside ACR

### DIFF
--- a/templates/default/dockerImage.sh.erb
+++ b/templates/default/dockerImage.sh.erb
@@ -46,6 +46,30 @@ elif [ "$1" == "gc" ] ; then
     docker exec registry bin/registry garbage-collect /etc/docker/registry/config.yml -m
 	  docker container restart registry
 
+<% else -%>
+
+elif [ "$1" == "delete-acr" ] ; then
+  registry=$2
+  image_prefix=$3
+  if [[ -z "$registry" ||  -z "$image_prefix" ]] then
+    echo ""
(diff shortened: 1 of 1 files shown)


templates/default/dockerImage.sh.erb
```diff
@@ -46,6 +46,27 @@ elif [ "$1" == "gc" ] ; then
     docker exec registry bin/registry garbage-collect /etc/docker/registry/config.yml -m
 	  docker container restart registry
 
+elif [ "$1" == "delete-acr" ] ; then
```
maismail, line 49: This should be under the else of this condition when `managed_docker_registry` is true 
https://github.com/logicalclocks/hopsworks-chef/pull/570/files#diff-944e67fb3fc477da8e732aeac8d41b01b35b98af540418d278610340363d1454R30

maismail (changes requested):

templates/default/dockerImage.sh.erb
lovew-lc: ahhh I read if managed=true. fixed in last commit

maismail: The block was correct you just need to add `<% else -%>` before it

lovew-lc: Ah sorry, I am not awake today.. fixed now

maismail: no worries

maismail (approved):

PR #584, opened 2021-03-15
[HOPSWORKS-2433] add Azure CA cert

ErmiasG:


recipes/letsencrypt.rb
```diff
@@ -14,6 +14,11 @@
 EOF
 end
 
+remote_file '/tmp/DigiCertGlobalRootG2.crt' do
```
kouzant, line 17: If you check a few lines above you'll see that this recipe is only for `hops.site` It's not included anywhere else. This is not the correct place.

kouzant (changes requested):

providers/certs.rb
```diff
@@ -175,6 +175,13 @@
     end
   end
 
+  # Download azure root CA cert
+  remote_file '/tmp/DigiCertGlobalRootG2.crt' do
```
kouzant, line 179: Could you please create another `action` to download and import Azure CA?

In a week from now we won't even remember we are abusing the action called `generate_int_certs`

providers/certs.rb
```diff
@@ -190,7 +197,9 @@
         # Import into the keystore
         keytool -importkeystore -destkeystore keystore.jks -srckeystore cert_and_key.p12 -srcstoretype PKCS12 -alias internal -srcstorepass #{node['hopsworks']['master']['password']} -deststorepass #{node['hopsworks']['master']['password']} -destkeypass #{node['hopsworks']['master']['password']}
 
-         keytool -import -noprompt -alias internal -file #{node['certs']['dir']}/certs/ca.cert.pem -keystore cacerts.jks -storepass #{node['hopsworks']['master']['password']}
+        keytool -import -noprompt -alias internal -file #{node['certs']['dir']}/certs/ca.cert.pem -keystore cacerts.jks -storepass #{node['hopsworks']['master']['password']}
+        
+        keytool -import -noprompt -alias digicertglobalrootg2 -file /tmp/DigiCertGlobalRootG2.crt -keystore cacerts.jks -storepass #{node['hopsworks']['master']['password']}
```
kouzant, line 202: Same as above, create a separate `action` and call it from `hopsworks::default`?

kouzant (changes requested):

kouzant (approved):

PR #588, opened 2021-03-23
[HOPSWORKS-2458] Online FS user

moritzmeister:


recipes/default.rb
```diff
@@ -221,6 +223,9 @@
   node.override['hopsworks']['reserved_project_names'] = "#{node['hopsworks']['reserved_project_names']},cloud"
 end
 
+# encrypt onlinefs user password
+encrypted_onlinfefs_password = Digest::SHA256.hexdigest node['onlinefs']['hopsworks']['password'] + node['onlinefs']['hopsworks']['salt']
```
jimdowling, line 227: I assume that the salt needed outside this recipe/cookbook - it's generated externally and stored in a table?
If not, why not just generate a random number?

recipes/default.rb
```diff
@@ -246,7 +251,10 @@
          :public_ip => public_ip,
          :dela_ip => dela_ip,
          :krb_ldap_auth => node['ldap']['enabled'].to_s == "true" || node['kerberos']['enabled'].to_s == "true",
-         :hops_version => get_hops_version(node['hops']['version'])
+         :hops_version => get_hops_version(node['hops']['version']),
+         :onlinefs_password => encrypted_onlinfefs_password,
+         :onlinefs_email => node['onlinefs']['hopsworks']['password'],
```
jimdowling, line 256: should this not be:
node['onlinefs']['hopsworks']['email']
?

templates/default/sql/dml/2.2.0.sql.erb
```diff
@@ -35,3 +35,8 @@ REPLACE INTO `hopsworks`.`variables`(`id`, `value`) VALUES ("fs_job_activity_tim
 
 REPLACE INTO `hopsworks`.`variables`(`id`, `value`) VALUES ("pypi_indexer_timer_interval", "<%= node['hopsworks']['pypi_indexer_timer_interval'] %>");
 REPLACE INTO `hopsworks`.`variables`(`id`, `value`) VALUES ("pypi_simple_endpoint", "<%= node['hopsworks']['pypi_simple_endpoint'] %>");
+
+INSERT INTO `hopsworks`.`users`
+(`username`, `password`, `email`, `fname`, `lname`, `activated`, `title`, `orcid`, `false_login`, `isonline`, `secret`, `validation_key`, `mode`, `password_changed`, `notes`, `status`, `max_num_projects`, `two_factor`, `tours_state` , `salt`) VALUES
+('onlinefs', '<%= @onlinefs_password %>', '<%= @onlinefs_email %>', 'OnlineFS', 'Server', '2021-03-16 16:17:00', 'Mr', '1234-5678-1234-5678', -1, 0, 'V3WBPS4G2WMQ53VA', NULL, 0, '2021-03-16 16:17:00', NULL, 2, 30, 0, 3, '<%= @onlinefs_salt %>')
+ON DUPLICATE KEY UPDATE `username`="onlinefs";
```
jimdowling, line 42: Is there a single db user called 'onlinefs'?
Before we had  an online database per online feature store. You would have a different username/password for each online FS. Is this a single global onlinefs user?

jimdowling (changes requested):

recipes/default.rb
SirOibaf, line 256: 2 points here:
- there is a typo here, it should be email not password
- If you are using `node` without any additional logic, you don't need to pass them as variables, you can just use the node object in the template. See below.

templates/default/sql/dml/2.2.0.sql.erb
```diff
@@ -35,3 +35,8 @@ REPLACE INTO `hopsworks`.`variables`(`id`, `value`) VALUES ("fs_job_activity_tim
 
 REPLACE INTO `hopsworks`.`variables`(`id`, `value`) VALUES ("pypi_indexer_timer_interval", "<%= node['hopsworks']['pypi_indexer_timer_interval'] %>");
 REPLACE INTO `hopsworks`.`variables`(`id`, `value`) VALUES ("pypi_simple_endpoint", "<%= node['hopsworks']['pypi_simple_endpoint'] %>");
+
+INSERT INTO `hopsworks`.`users`
+(`username`, `password`, `email`, `fname`, `lname`, `activated`, `title`, `orcid`, `false_login`, `isonline`, `secret`, `validation_key`, `mode`, `password_changed`, `notes`, `status`, `max_num_projects`, `two_factor`, `tours_state` , `salt`) VALUES
+('onlinefs', '<%= @onlinefs_password %>', '<%= @onlinefs_email %>', 'OnlineFS', 'Server', '2021-03-16 16:17:00', 'Mr', '1234-5678-1234-5678', -1, 0, 'V3WBPS4G2WMQ53VA', NULL, 0, '2021-03-16 16:17:00', NULL, 2, 30, 0, 3, '<%= @onlinefs_salt %>')
```
SirOibaf, line 41: Replace `@onlinefs_email` with `node['onlinefs']['hopsworks']['email']` and `@onlinefs_salt` with `node['onlinefs']['hopsworks']['salt']`

SirOibaf (changes requested):

recipes/default.rb
moritzmeister: 1. good catch, thanks
2. didn't know, will replace

moritzmeister: 1. good catch, thanks
2. didn't know, will replace

templates/default/sql/dml/2.2.0.sql.erb
moritzmeister: fixed

recipes/default.rb
moritzmeister: The reason we do this is because we need the clear text password in the onlinefs::default recipe to be able to log in as the user and get an api key for the service. The salt could just be a random string, yes, I just thought it's good to have it defined together with the password in the attributes of onlinefs-chef, and not somewhere in the middle of a recipe.

moritzmeister: changed it to use random salt

jimdowling (approved):

templates/default/sql/dml/2.2.0.sql.erb
```diff
@@ -35,3 +35,10 @@ REPLACE INTO `hopsworks`.`variables`(`id`, `value`) VALUES ("fs_job_activity_tim
 
 REPLACE INTO `hopsworks`.`variables`(`id`, `value`) VALUES ("pypi_indexer_timer_interval", "<%= node['hopsworks']['pypi_indexer_timer_interval'] %>");
 REPLACE INTO `hopsworks`.`variables`(`id`, `value`) VALUES ("pypi_simple_endpoint", "<%= node['hopsworks']['pypi_simple_endpoint'] %>");
+
+INSERT INTO `hopsworks`.`users`
+(`username`, `password`, `email`, `fname`, `lname`, `activated`, `title`, `orcid`, `false_login`, `isonline`, `secret`, `validation_key`, `mode`, `password_changed`, `notes`, `status`, `max_num_projects`, `two_factor`, `tours_state` , `salt`) VALUES
+('<%= node['onlinefs']['user'] %>', '<%= @onlinefs_password %>', '<%= node['onlinefs']['hopsworks']['email'] %>', 'OnlineFS', 'Server', '2021-03-16 16:17:00', 'Mr', '1234-5678-1234-5678', -1, 0, 'V3WBPS4G2WMQ53VA', NULL, 0, '2021-03-16 16:17:00', NULL, 2, 30, 0, 3, '<%= @onlinefs_salt %>')
+ON DUPLICATE KEY UPDATE `username`="onlinefs";
+
+INSERT INTO `hopsworks`.`user_group` (`uid`, `gid`) SELECT `uid`, 1007 FROM `hopsworks`.`users` WHERE `username`='<%= node['onlinefs']['user'] %>';
```
SirOibaf, line 44: The DML files are run at every update, so the second time they are run this row will already exists. we need to handle that.

SirOibaf (changes requested):

templates/default/sql/dml/2.2.0.sql.erb
moritzmeister: changed it to `INSERT IGNORE INTO`

SirOibaf (approved):

PR #589, opened 2021-03-23
add login into acr for acr support

berthoug:


templates/default/dockerImage.sh.erb
```diff
@@ -99,6 +99,13 @@ elif [ "$1" == "export" ] ; then
     docker pull $2 > /dev/null 2>&1
     docker run --rm $2 conda env export -n theenv
 
+elif [ "$1" == "loginaz" ] ; then
```
maismail, line 102: you should add it in the same block after `delete-acr`
https://github.com/logicalclocks/hopsworks-chef/blob/master/templates/default/dockerImage.sh.erb#L71
It will be only added to the script when managed containers is enabled

maismail (changes requested):

templates/default/dockerImage.sh.erb
berthoug: ok

maismail (approved):

PR #593, opened 2021-04-01
[HOPSWORKS-2478] make it possible to set livy_session_startup_timeout_seconds in the jupyter configuration

berthoug:
…seconds in the jupyter configuration

### DIFF
--- a/attributes/default.rb
+++ b/attributes/default.rb
@@ -454,3 +454,6 @@ default['hopsworks']['hdfs']['storage_policy']['log']         = "HOT"
 default["hopsworks"]['check_nodemanager_status']              = "false"
 
 default['hopsworks']['azure-ca-cert']['download-url']         = "https://cacerts.digicert.com/DigiCertGlobalRootG2.crt"
+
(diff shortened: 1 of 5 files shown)


attributes/default.rb
```diff
@@ -453,3 +453,6 @@
 default["hopsworks"]['check_nodemanager_status']              = "false"
 
 default['hopsworks']['azure-ca-cert']['download-url']         = "https://cacerts.digicert.com/DigiCertGlobalRootG2.crt"
+
+#livy
+default['livy_startup_timeout']           = "240"
```
robzor92, line 458:
```suggestion
default['hopsworks']['livy_startup_timeout']           = "240"
```

robzor92: The dml uses ['hopsworks']['livy_startup_timeout'] so I assume you missed to add the hopsworks part in the attribute

robzor92 (changes requested):

robzor92 (approved):

PR #619, opened 2021-05-19
[HOPSWORKS-2375] Prefix for joining to avoid clashing

davitbzh:


files/default/sql/ddl/2.3.0__initial_tables.sql
```diff
@@ -1351,6 +1351,7 @@ CREATE TABLE `training_dataset_join` (
   `feature_group_commit_id` BIGINT(20) NULL,
   `type` tinyint(5) NOT NULL DEFAULT 0,
   `idx` int(11) NOT NULL DEFAULT 0,
+  `prefix` VARCHAR(100) NULL,
```
moritzmeister, line 1354: No need for `varchar(100)`, feature names can maximum be 63 characters, I doubt someone would add a prefix with 100 chars.

moritzmeister: and then you should also add some validation in hopsworks, that throws an error if users provide a longer prefix

moritzmeister (commented):
I think we can make it a bit shorter.

moritzmeister (approved):

moritzmeister (approved):

PR #628, opened 2021-06-08
Don't overwrite PATH and LD_LIBRARY_PATH when launching sklearn docker

SirOibaf:
cointainers

PATH and LD_LIBRARY_PATH are already defined correctly on the Docker
image. Moreover this cause issues if the host OS is different than
the Docker image OS.


templates/default/sklearn_flask_server.py.erb
```diff
@@ -1,109 +0,0 @@
-"""
```
kouzant, line 1: I don't think you should remove this file. It's templated in install.rb:678 and used by `/srv/hops/sbin/sklearn_serving.sh`

kouzant (changes requested):

templates/default/sklearn_flask_server.py.erb
SirOibaf: Oh true, I was grepping with the `.erb` extension and didn't find anything. Reverting, thanks.

kouzant (approved):

PR #631, opened 2021-06-16
[HOPSWORKS-2584] Increase Schema Registry database schema column size

moritzmeister:


moritzmeister:
@SirOibaf not sure if the undo is correct. it will cut off longer schemas, or do we just not add an undo for this?

SirOibaf (approved):

PR #643, opened 2021-07-22
[HOPSWORKS-2661] Sort Feature Group columns by hive metastore index

moritzmeister:


files/default/sql/ddl/3.0.0__initial_tables.sql
```diff
@@ -1392,6 +1392,7 @@ CREATE TABLE `on_demand_feature` (
   `primary_column` tinyint(1) NOT NULL DEFAULT '0',
   `description` varchar(10000) COLLATE latin1_general_cs,
   `type` varchar(1000) COLLATE latin1_general_cs NOT NULL,
+  `idx` int(11) NOT NULL DEFAULT 0,
```
tkakantousis, line 1395: Why do we need a new column for ordering the features? We can use the`id` which is automatically incremented.

tkakantousis (approved):

PR #658, opened 2021-08-27
[HOPSWORKS-2705] append fix hadoop group

jimdowling:


kouzant:
@jimdowling the intention of the snippets you changed is to create the `hdfs` group, not the `hadoop` group as you suggest.

PR #678, opened 2021-11-04
[HOPSWORKS-2721] Compliance rules improvements

tkakantousis:


files/default/sql/ddl/updates/2.5.0.sql
```diff
@@ -0,0 +1,3 @@
+ALTER TABLE `hopsworks`.`validation_rule` MODIFY COLUMN description VARCHAR(200) CHARACTER SET latin1 COLLATE latin1_general_cs DEFAULT NULL;
+ALTER TABLE `hopsworks`.`validation_rule` DROP INDEX `unique_validation_rule`;
+ALTER TABLE `hopsworks`.`validation_rule` ADD CONSTRAINT `unique_validation_rule` UNIQUE KEY (`name`);
```
robzor92, line 3: This migration may fail if we have entries in the table with the same `name`, but as this is the validation_rule description table populated by us it should not happen right?

tkakantousis: Correct, the tables is not accessible by users and it does not contain duplicates.

robzor92 (approved):

PR #692, opened 2021-11-17
[HOPSWORKS-2763] Opensearch 1.3.1

jimdowling:
https://logicalclocks.atlassian.net/browse/HOPSWORKS-2763


Berksfile
```diff
@@ -4,24 +4,24 @@ source 'https://supermarket.chef.io'
 metadata
 
 cookbook 'conda', github: "logicalclocks/conda-chef", branch: "master"
-cookbook 'kagent', github: "logicalclocks/kagent-chef", branch: "master"
+cookbook 'kagent', github: "logicalclocks/kagent-chef", branch: "opensearch"
```
o-alex, line 7: can you please reset this to master branch

files/default/sql/ddl/updates/undo/2.4.0__undo.sql
```diff
@@ -74,3 +74,5 @@ ALTER TABLE `hopsworks`.`jupyter_settings` DROP COLUMN `no_limit`;
 ALTER TABLE `hopsworks`.`oauth_login_state` MODIFY COLUMN `state` VARCHAR(128);
 
 ALTER TABLE `hopsworks`.`feature_group` DROP COLUMN `event_time`;
+
+DROP VIEW project_username;
```
o-alex, line 78: is this needed? if yes can we move it to 2.6.0

recipes/install.rb
```diff
@@ -890,3 +890,13 @@
   mode "770"
   action :create
 end
+
+# Add logo for opensearch-dashboards
```
o-alex, line 894: do we need this? does the rebranding work?

o-alex (changes requested):

Berksfile
jimdowling: Yep.

recipes/install.rb
jimdowling: removed

files/default/sql/ddl/updates/undo/2.4.0__undo.sql
jimdowling: had removed just before you saw this.

o-alex (approved):

PR #706, opened 2021-12-28
[HOPSWORKS-2828] Add predictor column and scikit-learn version

javierdlrm:


tkakantousis (approved):
What will happen if a user installs another scikit-learn version in their project?

javierdlrm:
If a user installs another sklearn version in the project environment, train a model and save it as .pickle file. It may work due to backward compatibility from the scikit-learn side, but KFServing will use `node['scikit-learn']['version']` which is currently `0.22.2.post1` as specified in conda-chef. I would say that, similarly to tensorflow, this is the 'tested' version and we don't guarantee different versions to work as expected.

PR #713, opened 2022-01-20
[HOPSWORKS-2853]-append id token default null

ErmiasG:


maismail (approved):

maismail (commented):
shouldn't you add DEFAULT to these statements as well
https://github.com/logicalclocks/hopsworks-chef/blob/master/files/default/sql/ddl/updates/2.5.0.sql#L42

maismail (changes requested):
.

maismail (approved):

PR #718, opened 2022-02-09
[HOPSWORKS-2971] Time Series Split for training datasets

davitbzh:


files/default/sql/ddl/updates/undo/2.6.0__undo.sql
```diff
@@ -0,0 +1 @@
+ALTER TABLE `hopsworks`.`training_dataset` DROP COLUMN `train_type`;
```
kennethmhc, line 1:
```suggestion
ALTER TABLE `hopsworks`.`training_dataset` DROP COLUMN `split_type `;
```

files/default/sql/ddl/2.6.0__initial_tables.sql
```diff
@@ -1321,6 +1321,7 @@ CREATE TABLE `training_dataset` (
                                     `query` TINYINT(1) NOT NULL DEFAULT '0',
                                     `coalesce` TINYINT(1) NOT NULL DEFAULT '0',
                                     `train_split` VARCHAR(63) COLLATE latin1_general_cs DEFAULT NULL,
+                                    `split_type` INT NOT NULL DEFAULT 0,
```
kennethmhc, line 1324: question: `XmlEnum` may the class to int instead of string?

PR #720, opened 2022-02-16
[HOPSWORKS-2998] Pass variable to indicate if TLS is enabled or not to git container

gibchikafa:


SirOibaf:
@gibchikafa  - PR looks ok. Just for the future, if you need to template an attribute, you don't need to use the `@` notation, you can just provide the attribute in the template itself.

SirOibaf (approved):

PR #732, opened 2022-03-02
[HOPSWORKS-3020-1] Add description column in serving table

javierdlrm:


robzor92 (approved):

files/default/sql/ddl/2.6.0__initial_tables.sql
```diff
@@ -1165,6 +1165,7 @@ CREATE TABLE `serving` (
                            `project_id` int(11) NOT NULL,
                            `created` timestamp NOT NULL DEFAULT CURRENT_TIMESTAMP,
                            `name` varchar(255) COLLATE latin1_general_cs NOT NULL,
+                           `description` varchar(10000) COLLATE latin1_general_cs DEFAULT NULL,
```
robzor92, line 1168: We could potentially bring down the max size here to something between 1000-5000 as we may add more fields in the future that could be large. I'm not sure of the max number of bytes per row in rondb at the moment but 10000 characters is alot. :)

robzor92 (changes requested):

files/default/sql/ddl/2.6.0__initial_tables.sql
javierdlrm: Changed to 1000

robzor92 (approved):

PR #736, opened 2022-03-14
[HOPSWORKS-2944] Implement keyword related endpoints

bubriks:


kennethmhc:
please modify the update and undo files as well.

bubriks:
> please modify the update and undo files as well.

yes, it is pushed now

PR #737, opened 2022-03-16
[HOPSWORKS-2941] Training Dataset Resource

kennethmhc:


files/default/sql/ddl/updates/undo/2.6.0__undo.sql
```diff
@@ -1,11 +1,23 @@
...
 ALTER TABLE `hopsworks`.`feature_store_activity` DROP FOREIGN KEY `fsa_feature_view_fk`, DROP COLUMN feature_view_id;
 ALTER TABLE `hopsworks`.`training_dataset` DROP COLUMN `start_time`, DROP COLUMN `end_time`;
 DROP TABLE IF EXISTS `hopsworks`.`feature_view`;
 
+ALTER TABLE `hopsworks`.`training_dataset` DROP FOREIGN KEY `hopsfs_training_dataset_fk`;
+ALTER TABLE `hopsworks`.`training_dataset` ADD CONSTRAINT `hopsfs_training_dataset_fk`
+    FOREIGN KEY (`hopsfs_training_dataset_id`) REFERENCES `hopsfs_training_dataset` (`id`)
+        ON DELETE SET NULL ON UPDATE NO ACTION;
+
+ALTER TABLE `hopsworks`.`training_dataset` DROP FOREIGN KEY `FK_656_817`;
```
bubriks, line 15: maybe i didnt get it, but to me, it looks like the fk "FK_656_817" is dropped and then the same is created again.

bubriks: what i mean is that there doesn't seem to be any change between versions for this constraint

kennethmhc: It is because `CONSTRAINT name_version` depends on `feature_store_id` and it can only be dropped if `FK_656_817` had dropped.

kennethmhc: @bubriks

bubriks (approved):

PR #739, opened 2022-03-18
[HOPSWORKS-3052] Add a flag to skip namespace creation when creating …

maismail:
…a project
https://logicalclocks.atlassian.net/browse/HOPSWORKS-3052


attributes/default.rb
```diff
@@ -482,3 +482,5 @@
 default['hopsworks']['git_command_timeout_minutes']  = "60"
 
 default['hopsworks']['debug']  = "false"
+
+default['hopsworks']['kubernetes']['skip_namespace_creation']        = "false"
```
kouzant, line 486: This should be exported in metadata.rb

kouzant (changes requested):

kouzant (approved):

PR #742, opened 2022-03-29
[HOPSWORKS-3069] Allow StreamFeatureGroup in backend to be available offline only

moritzmeister:


files/default/sql/ddl/2.6.0__initial_tables.sql
```diff
@@ -1849,6 +1849,7 @@ CREATE TABLE IF NOT EXISTS `cached_feature_group` (
 CREATE TABLE IF NOT EXISTS `stream_feature_group` (
                                                       `id`                             INT(11) NOT NULL AUTO_INCREMENT,
                                                       `offline_feature_group`          BIGINT(20) NOT NULL,
+                                                      `online_enabled`                 TINYINT(1) NOt NULL DEFAULT 1,
```
davitbzh, line 1852: not sure if this is important but I think its `NOT NULL`

files/default/sql/ddl/updates/2.6.0.sql
```diff
@@ -78,6 +78,7 @@ ALTER TABLE `hopsworks`.`serving` ADD COLUMN `description` varchar(1000) COLLATE
 CREATE TABLE IF NOT EXISTS `stream_feature_group` (
                                                       `id`                             INT(11) NOT NULL AUTO_INCREMENT,
                                                       `offline_feature_group`          BIGINT(20) NOT NULL,
+                                                      `online_enabled`                 TINYINT(1) NOt NULL DEFAULT 1,
```
davitbzh, line 81: the same here, `NOT NULL`?

davitbzh (approved):

files/default/sql/ddl/updates/2.6.0.sql
moritzmeister: fixed

PR #752, opened 2022-05-05
[HOPSWORKS-3125] Feature view dev

kennethmhc:


files/default/sql/ddl/2.6.0__initial_tables.sql
```diff
@@ -1348,8 +1352,9 @@ CREATE TABLE `training_dataset` (
                                     `feature_view_id` INT(11) NULL,
                                     `start_time` TIMESTAMP NULL,
                                     `end_time` TIMESTAMP NULL,
+                                    `sample_ratio` FLOAT NULL,
                                     PRIMARY KEY (`id`),
-                                    UNIQUE KEY `name_version` (`feature_store_id`, `name`, `version`),
+                                    UNIQUE KEY `name_version` (`feature_store_id`, `feature_view_id`, `name`, `version`),
```
jimdowling, line 1357: Does TD name have to be part of the unique ID? 
I thought TD ids were unique for a FV, so name is not needed here?

jimdowling (approved):
Looks good, just one question about whether 'name' is needed in the unique key.

files/default/sql/ddl/2.6.0__initial_tables.sql
kennethmhc: It needs to be a part of the unique key at least before migration because existing TD do not have fv id.

PR #755, opened 2022-05-06
[HOPSWORKS-2986] Data validation v2

moritzmeister:


files/default/sql/ddl/2.6.0__initial_tables.sql
```diff
@@ -2270,3 +2270,65 @@ CREATE TABLE IF NOT EXISTS `git_repository_remotes` (
...
+    `feature_group_id` INT(11) NOT NULL,
+    `name` VARCHAR(63) NOT NULL,
+    `meta` VARCHAR(1000) DEFAULT "{}",
+    `data_asset_type` VARCHAR(50),
+    `ge_cloud_id` VARCHAR(200),
+    `run_validation` BOOLEAN DEFAULT TRUE,
+    `validation_ingestion_policy` VARCHAR(25),
+    `created_at` timestamp DEFAULT CURRENT_TIMESTAMP,
+    PRIMARY KEY (`id`),
+    CONSTRAINT `feature_group_suite_fk` FOREIGN KEY (`feature_group_id`) REFERENCES `feature_group` (`id`) ON DELETE CASCADE ON UPDATE NO ACTION
```
jimdowling, line 2285: So, expectation suites are tightly coupled to FGs.
That's ok, we should make it clear in the docs, though. 
Because we had previously said they are reusable over FGs.

files/default/sql/ddl/2.6.0__initial_tables.sql
```diff
@@ -2270,3 +2270,65 @@ CREATE TABLE IF NOT EXISTS `git_repository_remotes` (
...
+) ENGINE=ndbcluster DEFAULT CHARSET=latin1 COLLATE=latin1_general_cs;
+
+CREATE TABLE IF NOT EXISTS `expectation` (
```
jimdowling, line 2288: This pattern of having a row per expectation, inside an expectation suite, where the 
expectations are not reused over expectation_suites, will mean it is slow to retrieve
100s of expectations. I will not be surprised if we have to refactor to add all expectations as a 
JSON column inside the expectation suite. 
However, for now, we can let it go.

files/default/sql/ddl/2.6.0__initial_tables.sql
```diff
@@ -2270,3 +2270,65 @@ CREATE TABLE IF NOT EXISTS `git_repository_remotes` (
...
+    CONSTRAINT `feature_group_report_fk` FOREIGN KEY (`feature_group_id`) REFERENCES `feature_group` (`id`) ON DELETE CASCADE ON UPDATE NO ACTION,
+    CONSTRAINT `inode_result_fk` FOREIGN KEY (`inode_pid`,`inode_name`,`partition_id`) REFERENCES `hops`.`hdfs_inodes` (`parent_id`,`name`,`partition_id`) ON DELETE CASCADE ON UPDATE NO ACTION
+) ENGINE=ndbcluster DEFAULT CHARSET=latin1 COLLATE=latin1_general_cs;
+
+CREATE TABLE IF NOT EXISTS `validation_result` (
+    `id` INT(11) NOT NULL AUTO_INCREMENT,
+    `validation_report_id` INT(11) NOT NULL,
+    `expectation_id` INT(11) NOT NULL,
+    `success` BOOLEAN NOT NULL,
+    `result` VARCHAR(1000) NOT NULL,
```
jimdowling, line 2327: Is 1000 chars enough? Wouldn't hurt to raise it to 5 or 10k?

files/default/sql/ddl/2.6.0__initial_tables.sql
```diff
@@ -2270,3 +2270,65 @@ CREATE TABLE IF NOT EXISTS `git_repository_remotes` (
...
+    `meta` VARCHAR(1000) DEFAULT "{}",
+    `expectation_config` VARCHAR(2150) NOT NULL,
+    `exception_info` VARCHAR(1000) DEFAULT "{}",
```
jimdowling, line 2330: Or maybe this will be longer than 1k chars?
Wouldn't hurt to raise it to 5 or 10k?
Remember, total max size per row is ~30k.

files/default/sql/ddl/updates/2.6.0.sql
```diff
@@ -139,3 +139,67 @@ ALTER TABLE `hopsworks`.`jupyter_settings`
...
     DROP COLUMN git_config_id;
 DROP TABLE IF EXISTS `hopsworks`.`jupyter_git_config`;
+
+-- Data Validation Great Expectation
+CREATE TABLE IF NOT EXISTS `expectation_suite` (
+    `id` INT(11) NOT NULL AUTO_INCREMENT,
+    `feature_group_id` INT(11) NOT NULL,
+    `name` VARCHAR(63) NOT NULL,
+    `meta` VARCHAR(1000) DEFAULT "{}",
+    `data_asset_type` VARCHAR(50),
```
jimdowling, line 149: is 50 chars enough?

files/default/sql/ddl/2.6.0__initial_tables.sql
```diff
@@ -2270,3 +2270,65 @@ CREATE TABLE IF NOT EXISTS `git_repository_remotes` (
                                           KEY `git_repository_fk` (`repository`),
                                           CONSTRAINT `git_repository_fk` FOREIGN KEY (`repository`) REFERENCES `git_repositories` (`id`) ON DELETE CASCADE
 ) ENGINE=ndbcluster AUTO_INCREMENT=6164 DEFAULT CHARSET=latin1 COLLATE=latin1_general_cs;
+
+CREATE TABLE IF NOT EXISTS `expectation_suite` (
+    `id` INT(11) NOT NULL AUTO_INCREMENT,
+    `feature_group_id` INT(11) NOT NULL,
+    `name` VARCHAR(63) NOT NULL,
+    `meta` VARCHAR(1000) DEFAULT "{}",
+    `data_asset_type` VARCHAR(50),
```
jimdowling, line 2279: is 50 chars enough?

jimdowling (commented):
Couple of questions related to varchar column sizes.

files/default/sql/ddl/2.6.0__initial_tables.sql
moritzmeister: The idea is to allow multiple expectation suites per feature gorups in the future, so you can have one suite for data validation, one for drift monitoring etc.

moritzmeister: The previous reusability was a joke. cause you couldn't change the parameters of an expectation when reusing it and reapplying to another feature group. Also it was tightly coupled to the feature name. Now you can get the GE native object in the client, change it however you want and reuse it that way. This is possible because we always return GE objects instead of our own.

files/default/sql/ddl/2.6.0__initial_tables.sql
moritzmeister: This is faster than having one big json saved on HopsFS and having to deserialize it.

files/default/sql/ddl/2.6.0__initial_tables.sql
moritzmeister: it doesn't contain the full stack trace, just some summary. so it should be enough. you can always download the full report otherwise.

templates/default/sql/dml/2.6.0.sql.erb
```diff
@@ -20,4 +20,58 @@ REPLACE INTO `hopsworks`.`variables`(`id`, `value`) VALUES ("quotas_model_deploy
...
+INSERT INTO `hopsworks`.`great_expectation` (expectation_type, kwargs_template) VALUES ('expect_table_columns_to_match_set', '{"include_config": true, "catch_exceptions": false, "result_format": "BASIC", "column_set": null, "exact_match": true}');
+INSERT INTO `hopsworks`.`great_expectation` (expectation_type, kwargs_template) VALUES ('expect_table_row_count_to_be_between', '{"include_config": true, "catch_exceptions": false, "result_format": "BASIC", "min_value": null, "max_value": null, "meta": null}');
+INSERT INTO `hopsworks`.`great_expectation` (expectation_type, kwargs_template) VALUES ('expect_table_row_count_to_equal', '{"include_config": true, "catch_exceptions": false, "result_format": "BASIC", "value": null, "meta": null}');
+INSERT INTO `hopsworks`.`great_expectation` (expectation_type, kwargs_template) VALUES ('expect_table_row_count_to_equal_other_table', '{"include_config": true, "catch_exceptions": false, "result_format": "BASIC", "other_table_name": null}');
```
SirOibaf, line 77: I think we need to write a better query. This is going to be executed every time it runs. We had a similar issue with the Deequ rules where it ended up either failing or creating more rules with the same value.

Can you check you can re-run the same `INSERT` twice without failing?

moritzmeister: aha with every time you mean with upgrades?

moritzmeister: changed it to replace into

SirOibaf (approved):

PR #760, opened 2022-05-17
[HOPSWORKS-3109] remove aws_instance_role

DhananjayMukhedkar:


SirOibaf (changes requested):
Could you please also make a PR to the conda-chef repo to remove the attribute from:
- here: https://github.com/logicalclocks/conda-chef/blob/d48173fd7e01f0c1ce2858f27cfd3f90ff8e15ff/attributes/default.rb#L18
- here: https://github.com/logicalclocks/conda-chef/blob/d48173fd7e01f0c1ce2858f27cfd3f90ff8e15ff/metadata.rb#L124

Also we would need to remove the attribute from the 1.0.0 files on master, otherwise is going to fail:
- https://github.com/logicalclocks/hopsworks-chef/blob/9aadb1bc1d1aec0b0031b18ff2b4d3c6e138a38f/templates/default/sql/dml/1.0.0.sql.erb#L21
- https://github.com/logicalclocks/hopsworks-chef/blob/9aadb1bc1d1aec0b0031b18ff2b4d3c6e138a38f/templates/default/sql/dml/undo/1.0.0__undo.sql.erb#L18

SirOibaf (approved):

PR #762, opened 2022-05-24
[Hopsworks 3158] add hopsworks_public_hostname 

DhananjayMukhedkar:


attributes/default.rb
```diff
@@ -174,6 +174,9 @@
 # Master encryption password
 default['hopsworks']['encryption_password']      = "adminpw"
 
+# hostname
+default['hopsworks']['hopsworks_public_host'] = node['fqdn']
```
jimdowling, line 178: This won't always work. The node['FQDN'] is sometimes the private IP address on cloud VMs, such as GCP. It is basically the same as calling 'hostname' in bash. 
However, this is not a big problem, as you can ovveride node['hopsworks']['hopsworks_public_host'] in a Karamel cluster defn, because you put it in the metadata.rb file - good!
I  can change it in the hopsworks-installer.sh to override this chef attribute when installing. So it is ok.  Good to know about node['FQDN'], though.

DhananjayMukhedkar: Thanks!

moritzmeister (approved):

PR #769, opened 2022-06-08
[HOPSWORKS-3185] cascade api key scopes on api key delete

ErmiasG:


ErmiasG:
https://github.com/logicalclocks/hops-testing-ee/pull/539

PR #772, opened 2022-06-20
[HOPSWORKS-3209] Extend to avoid user encountering an error

vatj:


moritzmeister (changes requested):
You have to add the change also to the initial table definition.

moritzmeister (approved):

PR #775, opened 2022-07-13
[HOPSWORKS-3249] Separate sklearn from custom python deployments

javierdlrm:


robzor92 (changes requested):
We need to move this SQL migration code and expat to version 3.1

templates/default/expat-site.xml.erb
```diff
@@ -145,5 +145,6 @@
         io.hops.hopsworks.expat.migrations.featurestore.storageconnectors.StorageConnectorMigration
         io.hops.hopsworks.expat.migrations.featurestore.trainingdataset.CreateExternalTrainingDatasetInode
         io.hops.hopsworks.expat.migrations.featurestore.featureview.CreateFeatureViewFromTrainingDataset
+        io.hops.hopsworks.expat.migrations.serving.ServingModelFrameworkMigration
```
robzor92, line 148: This migration should be in version-31 I think?

robzor92 (approved):

PR #784, opened 2022-07-28
[HOPSWORKS-3264] Change the visibility of docker_cgroup_enabled variable

gibchikafa:


robzor92:
This PR should be the same as: https://github.com/logicalclocks/hopsworks-chef/pull/783

robzor92 (changes requested):
This PR should be the same as: https://github.com/logicalclocks/hopsworks-chef/pull/783

robzor92 (approved):

PR #785, opened 2022-08-02
[HOPSWORKS-3262] Add serving resource configuration variables

javierdlrm:


files/default/sql/ddl/3.1.0__initial_tables.sql
```diff
@@ -1177,8 +1177,8 @@ CREATE TABLE `serving` (
                            `local_dir` varchar(255) COLLATE latin1_general_cs DEFAULT NULL,
                            `batching_configuration` varchar(255) COLLATE latin1_general_cs DEFAULT NULL,
                            `optimized` tinyint(4) NOT NULL DEFAULT '0',
-                           `instances` int(11) NOT NULL DEFAULT '0',
-                           `transformer_instances` int(11) DEFAULT NULL,
+                           `instances` FLOAT NOT NULL,
```
robzor92, line 1180: Why are we changing instances and transformer_instances to FLOAT? In hopsworks-ee PR they are still of Integer type and that makes more sense since why would an instances field have a the fractional number?

robzor92 (changes requested):

files/default/sql/ddl/3.1.0__initial_tables.sql
javierdlrm: You are right, I forgot to revert those changes. Fixed.

robzor92 (approved):

templates/default/sql/dml/undo/3.0.0__undo.sql.erb
```diff
@@ -32,3 +32,11 @@ DELETE FROM `hopsworks`.`variables` WHERE `id`='docker_cgroup_monitor_interval';
 DELETE FROM `hopsworks`.`variables` WHERE `id`='prometheus_port';
 
 DELETE FROM `hopsworks`.`variables` WHERE `id`='kube_kserve_tensorflow_version';
+
+REPLACE INTO `hopsworks`.`variables` VALUES ('kube_docker_cores_fraction', '<%= node['kube-hops']['docker_cores_fraction'] %>', 1, 0);
```
robzor92, line 36: Since this attribute `'<%= node['kube-hops']['docker_cores_fraction'] %>'` does not exists after the kube-hops-chef PR is merged. We should replace it with '1.0' which is the old default value.

robzor92 (changes requested):

templates/default/sql/dml/undo/3.0.0__undo.sql.erb
javierdlrm: Fixed

robzor92 (approved):

PR #786, opened 2022-08-02
[HOPSWORKS-3262] Add serving resource configuration variables

javierdlrm:


robzor92 (approved):

templates/default/sql/dml/undo/3.0.0__undo.sql.erb
```diff
@@ -32,3 +32,11 @@ DELETE FROM `hopsworks`.`variables` WHERE `id`='docker_cgroup_monitor_interval';
 DELETE FROM `hopsworks`.`variables` WHERE `id`='prometheus_port';
 
 DELETE FROM `hopsworks`.`variables` WHERE `id`='kube_kserve_tensorflow_version';
+
+REPLACE INTO `hopsworks`.`variables` VALUES ('kube_docker_cores_fraction', '<%= node['kube-hops']['docker_cores_fraction'] %>', 1, 0);
```
robzor92, line 36: Since this attribute `'<%= node['kube-hops']['docker_cores_fraction'] %>' does not exists after the kube-hops-chef PR is merged. We should replace it with '1.0' which is the old default value.

robzor92 (changes requested):

templates/default/sql/dml/undo/3.0.0__undo.sql.erb
javierdlrm: Fixed

robzor92 (approved):

PR #801, opened 2022-09-15
[HOPSWORKS-3351] Add validation_time column in validation result table

vatj:


moritzmeister (changes requested):
Missing the undo statements.

PR #816, opened 2022-10-11
[FSTORE-326] update GCS, BIGQUERY, KAFKA connector inode FK 

DhananjayMukhedkar:
- Fixes storage connector auto-delete if key files are updated.
- JIRA https://hopsworks.atlassian.net/browse/FSTORE-326


SirOibaf:
Why solving the issue like this? Why can't we re-order the operations so that they are:
- write new credentials files
- commit update to storage connector
- remove old file

DhananjayMukhedkar:
The issue here was that the file was getting deleted at the time of file upload from UI which updates the Inode table and before saving  the updated connector. During the update call the connector was already deleted so I had to update the table foreign keys on Inode table. But may be there is better way of fixing it. 
Also, there is another PR for deleting key files for connectors https://github.com/logicalclocks/hopsworks-ee/pull/1124

moritzmeister (approved):

PR #818, opened 2022-10-13
[FSTORE-359] Rename Transformation Function Output Types

tdoehmen:
s. https://hopsworks.atlassian.net/browse/FSTORE-359

### DIFF
--- a/files/default/sql/ddl/updates/3.1.0.sql
+++ b/files/default/sql/ddl/updates/3.1.0.sql
@@ -69,6 +69,7 @@ DROP TABLE `dela`;
 DROP TABLE `hopssite_cluster_certs`;
 
 -- rename transformation functions output types
+SET SQL_SAFE_UPDATES = 0;
 UPDATE transformation_function
 SET output_type = 'STRING'
 WHERE output_type = 'StringType()';
@@ -112,3 +113,4 @@ WHERE output_type = 'DateType()';
 UPDATE transformation_function
 SET output_type = 'BOOLEAN'
 WHERE output_type = 'BooleanType()';
+SET SQL_SAFE_UPDATES = 1;
--- a/files/default/sql/ddl/updates/3.1.0.sql
+++ b/files/default/sql/ddl/updates/3.1.0.sql
@@ -67,3 +67,48 @@ CREATE TABLE IF NOT EXISTS `pki_serial_number` (
 
 DROP TABLE `dela`;
 DROP TABLE `hopssite_cluster_certs`;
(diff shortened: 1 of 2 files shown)


files/default/sql/ddl/updates/3.1.0.sql
```diff
@@ -67,3 +67,48 @@ CREATE TABLE IF NOT EXISTS `pki_serial_number` (
 
 DROP TABLE `dela`;
 DROP TABLE `hopssite_cluster_certs`;
+
+-- rename transformation functions output types
+UPDATE transformation_function
```
SirOibaf, line 72: This doesn't work. MySQL has an option called `safe_update` (see doc: https://dev.mysql.com/doc/refman/8.0/en/mysql-command-options.html#option_mysql_safe-updates) - where if you try to run a update statement without a `WHERE` condition on a key, it's going to fail.

It works if you try the update manually as you run it as root and safe update is disabled, but when we run the migration with flyway, we need to disable safe updates before doing this kind of migrations.

See example here: https://github.com/logicalclocks/hopsworks-chef/blob/f178c2b8a312b568fec9df4e4d8ce66ec969f38c/files/default/sql/ddl/updates/3.0.0.sql#L151

Also make sure you re-enable it afterwards.

files/default/sql/ddl/updates/undo/3.1.0__undo.sql
```diff
@@ -66,3 +66,49 @@ CREATE TABLE `hopssite_cluster_certs` (
                                           PRIMARY KEY (`cluster_name`)
 ) ENGINE=ndbcluster DEFAULT CHARSET=latin1 COLLATE=latin1_general_cs;
 /*!40101 SET character_set_client = @saved_cs_client */;
+
+
+-- rename transformation functions output types
+UPDATE transformation_function
```
SirOibaf, line 72: Same comment as above.

SirOibaf (changes requested):

files/default/sql/ddl/updates/3.1.0.sql
tdoehmen: Oh, thanks for the hint. It indeed worked just fine when I tested it manually.

files/default/sql/ddl/updates/undo/3.1.0__undo.sql
tdoehmen: done

SirOibaf (approved):

PR #823, opened 2022-10-18
[HWORKS-160] fix add glassfish to airflow group

ErmiasG:


recipes/install.rb
```diff
@@ -93,6 +93,14 @@
   not_if { node['install']['external_users'].casecmp("true") == 0 }
 end
 
+group node['airflow']['group'] do
+  action :modify
+  members ["#{node['hopsworks']['user']}"]  
+  append true
+  not_if { node['install']['external_users'].casecmp("true") == 0 }
+  only_if "getent group #{node['airflow']['group']}"
+end
```
SirOibaf, line 102: Karamel does not respect install recipes dependencies. You need to create the group in this recipe as well to make sure it is going to work no matter the execution order.

SirOibaf (changes requested):

SirOibaf (approved):

PR #833, opened 2022-11-06
[HWORKS-210] jupyter threads fill up executor pool

o-alex:
https://hopsworks.atlassian.net/browse/HWORKS-210

### DIFF
--- a/recipes/install.rb
+++ b/recipes/install.rb
@@ -267,6 +267,13 @@ node.override = {
               'maximumpoolsize' => 400,
               'taskqueuecapacity' => 20000,
               'description' => 'Hopsworks Conda Executor Service'
+          },
+          'concurrent/jupyterExecutorService' => {
+              'threadpriority' => 8,
+              'corepoolsize' => 300,
+              'maximumpoolsize' => 300,
+              'taskqueuecapacity' => 1000,
+              'description' => 'Hopsworks Jupyter Executor Service'
           }
         },
         'managed_scheduled_executor_services' => {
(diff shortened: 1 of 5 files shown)


recipes/install.rb
```diff
@@ -267,6 +267,13 @@
               'maximumpoolsize' => 400,
               'taskqueuecapacity' => 20000,
               'description' => 'Hopsworks Conda Executor Service'
+          },
+          'concurrent/jupyterExecutorService' => {
```
SirOibaf, line 271: This is problematic during the upgrade. When an upgrade runs this configuration doesn't get applied. You need to create the pool also in the default recipe.

recipes/install.rb
```diff
@@ -267,6 +267,13 @@
...
+              'threadpriority' => 8,
+              'corepoolsize' => 300,
+              'maximumpoolsize' => 300,
+              'taskqueuecapacity' => 1000,
```
SirOibaf, line 275: I think we would also benefit from making these parameters configurable.

SirOibaf (changes requested):

recipes/install.rb
SirOibaf: Also I thought we didn't want to set the queue for this pool since it doesn't make much sense to wait

o-alex: I will make the params configurable.
You need the queue to pause connections from tornado that do not have resources to run. If I put no queue, the connection will fail with an error, but tornado(jupyter) will eagerly try to reconnect... and we end up with reconnects every 100ms.

recipes/install.rb
o-alex: done

SirOibaf (approved):

PR #871, opened 2022-12-23
[HWORKS-357] Fix upgrade from Glassfish 4 to Glassfish 5

SirOibaf:


recipes/install.rb
```diff
@@ -28,6 +18,25 @@
   EOF
 end
 
+# When we upgrade to version 3.1.0 make sure we have set the subjects
+# of CAs
+if conda_helpers.is_upgrade && Gem::Version.new(node['install']['current_version']) < Gem::Version.new('3.1.0')
+  if node['hopsworks']['pki']['root']['name'].empty? || node['hopsworks']['pki']['intermediate']['name'].empty?
+    raise "It is an upgrade and Hopsworks CA subject name are not set. You have to set it to what was the previous names"
```
kouzant, line 25: Would it be better if we silently override the names with
root: "C=SE, ST=Sweden, L=Stockholm, O=SICS, CN=HopsRootCA"
intermediate: "C=SE, ST=Sweden, O=SICS, CN=HopsIntermediateCA"
kubernetes: "C=SE, ST=Sweden, L=Stockholm, O=LogicalClocks, CN=KubeHopsIntermediateCA"

They can't be different for versions < 3.1 and we won't have to set them for upgrades

recipes/install.rb
```diff
@@ -848,6 +857,24 @@
...
+    user 'root'
+    code <<-EOH
+      set -e
+      cp -p #{node['hopsworks']['domains_dir']}4/domain1/config/internal* #{node['hopsworks']['config_dir']}/
+      cp -p #{node['hopsworks']['config_dir']}/internal_bundle.crt #{node['hopsworks']['config_dir']}/certificate_bundle.pem
+
+      # Hardcoded because the attribute doesn't exists anymore
+      cp #{node['hopsworks']['dir']}/certs-dir/certs/ca.cert.pem #{node['hopsworks']['config_dir']}/root_ca.pem
+
+      cp -p #{node['hopsworks']['domains_dir']}4/flyway/sql/* #{theDomain}/flyway/sql/
```
kouzant, line 873: And also `chown #{node['hopsworks']['user']} /srv/hops/certs-dir/certs/ca.cert.pem`

kouzant (changes requested):

recipes/install.rb
SirOibaf: I don't think it's possible. Seems from the code that the configuration is then passed to the DML files. Which means that in all future updates we need to set those configurations, even if the current version is more than 3.1.0

recipes/install.rb
```diff
@@ -28,6 +18,25 @@
...
+    raise "It is an upgrade and Hopsworks CA subject name are not set. You have to set it to what was the previous names"
+  end
+
+  # To support migration from Payara 4 to Payara 5 we need to re-generate the domain.
+  # We can drop the existing domain, but we need to save some files from it
+  bash 'Move old domains dir to domains4 for backup' do
+    user 'root'
+    code <<-EOH
+      set -e
+      mv -f #{node['hopsworks']['domains_dir']} #{node['hopsworks']['domains_dir']}4
```
robzor92, line 34: The 4 here is intended right?

robzor92 (changes requested):

robzor92 (approved):

PR #872, opened 2023-01-02
[HWORKS-356] Redirect standard error to log file

javierdlrm:


SirOibaf (approved):

templates/default/zip-background.sh.erb
```diff
@@ -21,6 +21,7 @@ cd "$1"
 freespace=$(df "$PWD" | awk '/[0-9]%/{print $(NF-2)}')
 freespace=$((freespace * 1000))
 
+hdfsFile=$2
```
SirOibaf, line 24: Can you put this between quotes - `"$2"`

SirOibaf (changes requested):

templates/default/zip-background.sh.erb
javierdlrm: Fixed

SirOibaf (approved):

PR #878, opened 2023-01-25
[FSTORE-626] Draft tables for persistence

vatj:


files/default/sql/ddl/3.2.0__initial_tables.sql
```diff
@@ -2356,3 +2356,57 @@ CREATE TABLE IF NOT EXISTS `hdfs_command_execution` (
...
+    CONSTRAINT `fm_descriptive_stats_fk` FOREIGN KEY (`feature_monitoring_config_id`) REFERENCES `feature_monitoring_configuration` (`id`) ON DELETE CASCADE ON UPDATE NO ACTION
+) ENGINE = ndbcluster DEFAULT CHARSET = latin1 COLLATE = latin1_general_cs;
+
+CREATE TABLE IF NOT EXISTS `feature_monitoring_result` (
+    `id` INT(11) NOT NULL AUTO_INCREMENT,
+    `feature_monitoring_config_id` INT(11) NOT NULL,
+    `job_id` INT(11) NOT NULL, -- dummy this should become ref to another table
+    `execution_id` INT(11) NOT NULL,  -- dummy this should become ref to another table
+    `monitoring_time` timestamp DEFAULT CURRENT_TIMESTAMP,
+    `triggered_alert` BOOLEAN DEFAULT FALSE,
```
javierdlrm, line 2406: Just a reminder here, about other naming proposals for this field. Maybe `shifted`, `shift_detected`, `out_of_bounds`?@DhananjayMukhedkar

files/default/sql/ddl/3.2.0__initial_tables.sql
```diff
@@ -2356,3 +2356,57 @@ CREATE TABLE IF NOT EXISTS `hdfs_command_execution` (
...
+    `threshold` FLOAT,
+    `compare_on` INT(11) NOT NULL,
+    PRIMARY KEY (`id`),
+    CONSTRAINT `fm_descriptive_stats_fk` FOREIGN KEY (`feature_monitoring_config_id`) REFERENCES `feature_monitoring_configuration` (`id`) ON DELETE CASCADE ON UPDATE NO ACTION
+) ENGINE = ndbcluster DEFAULT CHARSET = latin1 COLLATE = latin1_general_cs;
+
+CREATE TABLE IF NOT EXISTS `feature_monitoring_result` (
+    `id` INT(11) NOT NULL AUTO_INCREMENT,
+    `feature_monitoring_config_id` INT(11) NOT NULL,
+    `job_id` INT(11) NOT NULL, -- dummy this should become ref to another table
```
javierdlrm, line 2403: Do we need `job_id` if we already have `execution_id`? I think we could move `job_id` to `feature_monitoring_configuration`. Since the job parameters do not change, we can reuse the same job and just trigger another execution in every statistics computation (i.e., same job_id, new execution_id).

files/default/sql/ddl/3.2.0__initial_tables.sql
DhananjayMukhedkar: Yes I would agree, `triggered_alert` could be confusing as a name

files/default/sql/ddl/3.2.0__initial_tables.sql
```diff
@@ -2356,3 +2356,57 @@ CREATE TABLE IF NOT EXISTS `hdfs_command_execution` (
...
 ) ENGINE=ndbcluster DEFAULT CHARSET=latin1 COLLATE=latin1_general_cs;
+
+CREATE TABLE IF NOT EXISTS `feature_monitoring_configuration` (
+    `id` INT(11) NOT NULL AUTO_INCREMENT,
+    `feature_group_id` INT(11),
+    `feature_view_id` INT(11),
+    `feature_name` VARCHAR(63) COLLATE latin1_general_cs NOT NULL,
+    `enabled` BOOLEAN DEFAULT TRUE,
+    `feature_monitoring_type` tinyint(4) NOT NULL,
+    `alert_config` VARCHAR(63) COLLATE latin1_general_cs NOT NULL, -- dummy this should become ref to another table
```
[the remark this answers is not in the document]
DhananjayMukhedkar: I agree, alerts config should be optional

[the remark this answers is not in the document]
vatj: Ja, it should. Fixing it

files/default/sql/ddl/3.2.0__initial_tables.sql
```diff
@@ -2356,3 +2356,57 @@ CREATE TABLE IF NOT EXISTS `hdfs_command_execution` (
...
+    `scheduler_config` VARCHAR(63) COLLATE latin1_general_cs NOT NULL, -- dummy this should become ref to another table
+    PRIMARY KEY (`id`),
+    KEY (`feature_name`),
+    CONSTRAINT `fg_monitoring_config_fk` FOREIGN KEY (`feature_group_id`) REFERENCES `feature_group` (`id`) ON DELETE CASCADE ON UPDATE NO ACTION,
+    CONSTRAINT `fv_monitoring_config_fk` FOREIGN KEY (`feature_view_id`) REFERENCES `feature_view` (`id`) ON DELETE CASCADE ON UPDATE NO ACTION
+) ENGINE = ndbcluster DEFAULT CHARSET = latin1 COLLATE = latin1_general_cs;
+
+CREATE TABLE IF NOT EXISTS `monitoring_window_builder` (
+    `id` INT(11) NOT NULL AUTO_INCREMENT,
+    `feature_monitoring_config_id` INT(11) NOT NULL,
```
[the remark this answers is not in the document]
vatj: Maybe, I am not sure from a SQL perspective which is faster. I think that approaches is less to store though as you need only a boolean here rather than two ints in the config, but that's nothing anyway

files/default/sql/ddl/3.2.0__initial_tables.sql
vatj: Ja choose one together that you think is the right one I will put it in accordingly

files/default/sql/ddl/3.2.0__initial_tables.sql
vatj: That's a good point I will make the change

PR #886, opened 2023-02-09
[FSTORE-628] add testconnector script

DhananjayMukhedkar:


templates/default/testconnector-launch.sh.erb
```diff
@@ -0,0 +1,27 @@
...
+  help
+fi
+
+REGISTRY_URL='<%= @registry_addr %>'
+FILENAME=$(basename "$1")
+FOLDER=$(dirname "$1")
+# TODO image name as parameter?
+IMAGE="$2"
+
+if docker run --rm -v $FOLDER:/tmp $REGISTRY_URL/$IMAGE:latest /tmp/$FILENAME; then
```
moritzmeister, line 21: we should probably have a directory in `/tmp/testconnector` or something.
But wait, we agreed in the design log to write it to the staging dir in a subdirectory, in the backend you have it right `/srv/hops/staging`?

templates/default/testconnector-launch.sh.erb
```diff
@@ -0,0 +1,27 @@
...
+    echo "usage: $0 [input_json_filepath image_name]"
+    echo ""
+    exit 1
+}
+
+if [ "$#" != "2" ] || [ "$1" == "-h" ] ; then
+  help
+fi
+
+REGISTRY_URL='<%= @registry_addr %>'
```
maismail, line 15: This won't work in the cloud where each user will have its own ECR/ACR docker registry.
We should instead do the same as in Git where the full docker image name is passed to the script 
https://github.com/logicalclocks/hopsworks-ee/blob/master/hopsworks-common/src/main/java/io/hops/hopsworks/common/git/util/GitContainerArgumentsWriter.java#L106

maismail (changes requested):

templates/default/testconnector-launch.sh.erb
DhananjayMukhedkar: fixed

templates/default/testconnector-launch.sh.erb
DhananjayMukhedkar: yes we are using the staging dir `/srv/hops/staging` as the work directory in hops. The `/tmp` was only for the container destination directory. Now I have changed that to `/home` as `/tmp is not created by default

templates/default/testconnector-launch.sh.erb
```diff
@@ -0,0 +1,27 @@
...
+FILENAME=$(basename "$1")
+FOLDER=$(dirname "$1")
+# TODO image name as parameter?
+IMAGE="$2"
+
+if docker run --rm -v $FOLDER:/tmp $REGISTRY_URL/$IMAGE:latest /tmp/$FILENAME; then
+   exit 0
+else
+   exit 9
```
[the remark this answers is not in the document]
DhananjayMukhedkar: I was using it in integration spec to identify failure response from container. I just needed a unique non-zero value (1 would also work). For internal failures like process time-outs we get a different error code.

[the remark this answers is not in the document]
moritzmeister: I am just unsecure cause 9 has some specific meaning. Also if you call `exit(code)` in the python script, do you actrualy need the if statement here, won't it get propagated?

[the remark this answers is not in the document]
DhananjayMukhedkar: removed the exit codes, now the same return code from the python script should be propogated

recipes/default.rb
```diff
@@ -1147,6 +1147,13 @@
   action :create
 end
 
+directory node['hopsworks']['data_volume']['staging_dir'] + "/connectors"  do
```
moritzmeister, line 1150: We need to handle upgrades also.

templates/default/testconnector-launch.sh.erb
```diff
@@ -0,0 +1,22 @@
...
+if [ "$#" != "2" ] || [ "$1" == "-h" ] ; then
+  help
+fi
+
+
+FILENAME=$(basename "$1")
+FOLDER=$(dirname "$1")
+IMAGE="$2"
+
+docker run --rm -v $FOLDER:/home $IMAGE /home/$FILENAME -v;
```
maismail, line 20: I suppose you assume here that the image is already in the registry, then you would need to do similar to git
https://github.com/logicalclocks/hops-hadoop-chef/commit/3a8f812819b7aa8658df26383ade33ba1c9f59c6

maismail (changes requested):

recipes/default.rb
DhananjayMukhedkar: reverted change and included in hopsworks backend

templates/default/sql/dml/3.2.0.sql.erb
```diff
@@ -3,3 +3,5 @@ REPLACE INTO `hopsworks`.`variables`(`id`, `value`, `visibility`) VALUES ("max_a
 REPLACE INTO `hopsworks`.`variables`(`id`, `value`) VALUES ("hdfs_file_op_job_util", "hdfs:///user/<%= node['hadoop_spark']['user'] %>/hdfs_file_operations-<%= node['hadoop_spark']['hopsworks_jobs_py']['version'] %>.py");
 
 REPLACE INTO `hopsworks`.`variables`(`id`, `value`) VALUES ("git_image_version", "<%= node['hops']['docker']['git']['image']['version'] %>");
+
+  REPLACE INTO `hopsworks`.`variables`(`id`, `value`) VALUES ("testconnector_image_version", "<%= node['hops']['docker']['testconnector']['image']['version'] %>");
```
moritzmeister, line 7:
```suggestion
REPLACE INTO `hopsworks`.`variables`(`id`, `value`) VALUES ("testconnector_image_version", "<%= node['hops']['docker']['testconnector']['image']['version'] %>");
```

templates/default/testconnector-launch.sh.erb
moritzmeister: @DhananjayMukhedkar are you addressing this?

moritzmeister (changes requested):

templates/default/testconnector-launch.sh.erb
moritzmeister: ah i saw you got a pr there.

DhananjayMukhedkar: yes, just trying to test this in a  new VM

moritzmeister (approved):

maismail (approved):

PR #898, opened 2023-03-14
[FSTORE-742] Statistics persistence in the DB

javierdlrm:


files/default/sql/ddl/3.2.0__initial_tables.sql
```diff
@@ -2364,4 +2364,28 @@ CREATE TABLE IF NOT EXISTS `feature_monitoring_result` (
     `difference` FLOAT,
     PRIMARY KEY (`id`),
     CONSTRAINT `config_monitoring_result_fk` FOREIGN KEY (`feature_monitoring_config_id`) REFERENCES `feature_monitoring_config` (`id`) ON DELETE CASCADE ON UPDATE NO ACTION
-) ENGINE = ndbcluster DEFAULT CHARSET = latin1 COLLATE = latin1_general_cs;
\ No newline at end of file
+) ENGINE = ndbcluster DEFAULT CHARSET = latin1 COLLATE = latin1_general_cs;
+
+CREATE TABLE IF NOT EXISTS `feature_descriptive_statistics` (
```
vatj, line 2369: Do you think there is value in splitting the stats for categorical and numerical columns into two distinct tables?

javierdlrm: At the moment I don't see the point on splitting them, since the stats we compute for categorical features are a subset of the ones computed for the numerical ones.

PR #902, opened 2023-03-16
[FSTORE-612] Feature monitoring alpha of the pre-pre-alpha

vatj:


files/default/sql/ddl/3.2.0__initial_tables.sql
```diff
@@ -2311,3 +2311,97 @@ CREATE TABLE IF NOT EXISTS `hdfs_command_execution` (
...
+    `percentiles` BLOB,
+    `approx_num_distinct_values` BIGINT NULL,
+    -- with exactUniqueness
+    `distinctness` FLOAT NULL,
+    `entropy` FLOAT NULL,
+    `uniqueness` FLOAT NULL,
+    `exact_num_distinct_values` BIGINT NULL,
+    -- for filtering
+    `start_time` TIMESTAMP NULL,
+    `end_time` TIMESTAMP NULL,
```
moritzmeister, line 2393: Is it possible to have an index over the filtering columns, so we don't have to do a whole table scan?

files/default/sql/ddl/3.2.0__initial_tables.sql
```diff
@@ -2276,4 +2276,4 @@ CREATE TABLE IF NOT EXISTS `hdfs_command_execution` (
   KEY `fk_hdfs_file_command_2_idx` (`src_inode_partition_id`,`src_inode_pid`,`src_inode_name`),
   CONSTRAINT `fk_hdfs_file_command_1` FOREIGN KEY (`execution_id`) REFERENCES `executions` (`id`) ON DELETE CASCADE,
   CONSTRAINT `fk_hdfs_file_command_2` FOREIGN KEY (`src_inode_partition_id`,`src_inode_pid`,`src_inode_name`) REFERENCES `hops`.`hdfs_inodes` (`partition_id`, `parent_id`, `name`) ON DELETE CASCADE
-) ENGINE=ndbcluster DEFAULT CHARSET=latin1 COLLATE=latin1_general_cs;
+) ENGINE=ndbcluster DEFAULT CHARSET=latin1 COLLATE=latin1_general_cs;
```
SirOibaf, line 2279: Please avoid changing old files.

files/default/sql/ddl/updates/3.2.0.sql
```diff
@@ -187,4 +187,3 @@ ALTER TABLE `hopsworks`.`feature_store_bigquery_connector`
   DROP COLUMN `key_inode_pid`,
   DROP COLUMN `key_inode_name`,
   DROP COLUMN `key_partition_id`;
-
```
SirOibaf, line 190: Same here.

files/default/sql/ddl/updates/3.3.0.sql
```diff
@@ -100,3 +100,229 @@ ALTER TABLE `hopsworks`.`transformation_function` DROP KEY `inode_fn_fk`;
...
+ALTER TABLE `hopsworks`.`feature_store_activity`
+    DROP FOREIGN KEY `fs_act_stat_fk`,
+    DROP COLUMN `statistics_id`;
+
+ALTER TABLE `hopsworks`.`feature_store_statistic`
+    DROP FOREIGN KEY `fg_fk_fss`,
+    DROP FOREIGN KEY `fg_ci_fk_fss`,
+    DROP FOREIGN KEY `td_fk_fss`,
+    DROP FOREIGN KEY `inode_fk`;
+DROP TABLE IF EXISTS `hopsworks`.`feature_store_statistic`;
```
SirOibaf, line 253: Why do we need to drop the FK if we are just dropping the entire table?

files/default/sql/ddl/updates/3.3.0.sql
```diff
@@ -100,3 +100,229 @@ ALTER TABLE `hopsworks`.`transformation_function` DROP KEY `inode_fn_fk`;
...
+
+-- Update feature_store_activity foreign keys
+ALTER TABLE `hopsworks`.`feature_store_activity`
+    ADD COLUMN `feature_group_statistics_id` INT(11) NULL,
+    ADD COLUMN `feature_view_statistics_id` INT(11) NULL,
+    ADD COLUMN `training_dataset_statistics_id` INT(11) NULL,
+    ADD CONSTRAINT `fs_act_fg_stat_fk` FOREIGN KEY (`feature_group_statistics_id`) REFERENCES `feature_group_statistics` (`id`) ON DELETE CASCADE ON UPDATE NO ACTION,
+    ADD CONSTRAINT `fs_act_fv_stat_fk` FOREIGN KEY (`feature_view_statistics_id`) REFERENCES `feature_view_statistics` (`id`) ON DELETE CASCADE ON UPDATE NO ACTION,
+    ADD CONSTRAINT `fs_act_td_stat_fk` FOREIGN KEY (`training_dataset_statistics_id`) REFERENCES `training_dataset_statistics` (`id`) ON DELETE CASCADE ON UPDATE NO ACTION;
+-- TODO: SQL Query populate ids based on statistics_id.
```
SirOibaf, line 243: Missing migration

metadata.rb
```diff
@@ -1004,6 +1004,20 @@
...
+          
+# Feature monitoring feature flag
+attribute "hopsworks/enable_feature_monitoring",
+          :description => "Whether to enable feature monitoring with scheduling and alerting",
+          :type => 'string'
+          
+# Feature monitoring feature flag
+attribute "hopsworks/enable_feature_monitoring",
+          :description => "Whether to enable feature monitoring with scheduling and alerting",
           :type => 'string'
```
SirOibaf, line 1021: Duplicate there

files/default/sql/ddl/updates/3.3.0.sql
SirOibaf: Additionally: 
- How do we migrate old statistics to the new format? 
- How do we handle feature group correlation and histograms?

SirOibaf (changes requested):

files/default/sql/ddl/3.2.0__initial_tables.sql
javierdlrm: The filtering column is feature_name. Start and end times were moved to a different table.

files/default/sql/ddl/updates/3.3.0.sql
javierdlrm: > Why do we need to drop the FK if we are just dropping the entire table?

Fixed.

> How do we migrate old statistics to the new format?

The idea is to implement an expat migration that reads the hdfs files, extracts and stores the descriptive statistics into the database, and if histogram or correlations are available, updates the hdfs file accordingly and stores the string path in the db.

For the feature store activity table, just SQL queries should work.

> How do we handle feature group correlation and histograms?

They are kept in hdfs files, which path is stored in a column in the feature descriptive statistics table.

files/default/sql/ddl/updates/3.3.0.sql
javierdlrm: Yes, I'll add it together with the expat migration.

files/default/sql/ddl/3.3.0__initial_tables.sql
```diff
@@ -2264,3 +2299,168 @@ CREATE TABLE IF NOT EXISTS `hdfs_command_execution` (
...
+    `mean` FLOAT,
+    `stddev` FLOAT,
+    `percentiles` BLOB,
+    -- with exactUniqueness
+    `distinctness` FLOAT,
+    `entropy` FLOAT,
+    `uniqueness` FLOAT,
+    `exact_num_distinct_values` BIGINT,
+    -- extended stats (hdfs file): histogram, correlations, kll
+    `extended_statistics_path` VARCHAR(255),
```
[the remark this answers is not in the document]
javierdlrm: We have 1 file per feature (if histograms, correlations or kll were computed), but not necessarily per commit.

PR #905, opened 2023-03-21
[FSTORE-749] Move OnlineFS Kafka offsets in RonDB

bubriks:


files/default/sql/ddl/3.2.0__initial_tables.sql
```diff
@@ -943,6 +943,21 @@ CREATE TABLE `project_topics` (
 ) ENGINE=ndbcluster DEFAULT CHARSET=latin1 COLLATE=latin1_general_cs;
 /*!40101 SET character_set_client = @saved_cs_client */;
 
+--
+-- Table structure for table `project_topic_offsets`
+--
+
+/*!40101 SET @saved_cs_client     = @@character_set_client */;
+/*!40101 SET character_set_client = utf8 */;
+CREATE TABLE `project_topic_offsets` (
```
SirOibaf, line 952: This is not a good idea. The technical direction of the product is to isolate components, here we are doing the opposite direction. We are making the OnlineFS metadata part of the Hopsworks metadata. 
 
What about we have a checkpoint table for each online feature store? Or at least have a separate database for the OnlineFS.

SirOibaf (changes requested):

files/default/sql/ddl/3.2.0__initial_tables.sql
bubriks: This table technically could be used by any process that requires the storage of Kafka offset (not only onlinefs), but I get why you want it separate. The problem is that related data would not be stored in a single place. Wouldn't it maybe be better to instead separate the whole Kafka part in its separate database (topics, schemas, offsets, etc.)?

PR #906, opened 2023-03-21
[FSTORE-742] Add foreign keys to FM config and result

javierdlrm:


javierdlrm:
There are no actual deletions. It is just that table creation needs to be reorder due to the foreign keys.

vatj (approved):

PR #908, opened 2023-03-27
[HWORKS-218] upload_staging_dir

ErmiasG:
https://github.com/logicalclocks/hops-testing-ee/pull/610


attributes/default.rb
```diff
@@ -67,6 +67,7 @@
 
 default['hopsworks']['staging_dir']              = node['hopsworks']['dir'] + "/staging"
 default['hopsworks']['conda_cache']              = node['hopsworks']['staging_dir'] + "/glassfish_conda_cache"
+default['hopsworks']['upload_staging_dir']       = "Projects"
```
SirOibaf, line 70: Why do we need a variable to say that the chunks are going to be placed into the project?

PR #912, opened 2023-03-30
[FSTORE-674][FSTORE-675] feature monitoring alert changes to feature_group_alert

DhananjayMukhedkar:


files/default/sql/ddl/updates/3.2.0.sql
```diff
@@ -136,3 +136,14 @@ CREATE TABLE IF NOT EXISTS `job_schedule` (
     UNIQUE KEY `job_id` (`job_id`),
     CONSTRAINT `fk_schedule_job` FOREIGN KEY (`job_id`) REFERENCES `jobs` (`id`) ON DELETE CASCADE
     ) ENGINE=ndbcluster DEFAULT CHARSET=latin1 COLLATE=latin1_general_cs;
+
+-- alert changes for feature monitoring
+ALTER TABLE `feature_group_alert`
+    MODIFY `feature_group_id` int NULL,
+    ADD `feature_view_id` int NULL,
+    ADD `entity_type` int NOT NULL,
```
vatj, line 144: You need to provide a default here for things that already exist in the db

vatj (commented):
Test the upgrade with data in the alert table please

files/default/sql/ddl/updates/3.2.0.sql
DhananjayMukhedkar: added default for entity type. For other FG id and FV id they are foreign keys, dont think we need anything there.

DhananjayMukhedkar: Tested update DDL with data in table, works as expected

files/default/sql/ddl/3.2.0__initial_tables.sql
```diff
@@ -1995,21 +1995,29 @@ CREATE TABLE IF NOT EXISTS `job_alert` (
...
-                                                     UNIQUE KEY `unique_feature_group_alert` (`feature_group_id`,`status`),
-                                                     KEY `fk_feature_group_alert_2_idx` (`feature_group_id`),
-                                                     KEY `fk_feature_group_alert_1_idx` (`receiver`),
-                                                     CONSTRAINT `fk_feature_group_alert_1` FOREIGN KEY (`receiver`) REFERENCES `alert_receiver` (`id`) ON DELETE CASCADE,
-                                                     CONSTRAINT `fk_feature_group_alert_2` FOREIGN KEY (`feature_group_id`) REFERENCES `feature_group` (`id`) ON DELETE CASCADE
-) ENGINE=ndbcluster DEFAULT CHARSET=latin1 COLLATE=latin1_general_cs;
+CREATE TABLE IF NOT EXISTS `hopsworks`.`feature_group_alert` (
+                                                             `id` int AUTO_INCREMENT
+                                                                 PRIMARY KEY,
+                                                             `feature_group_id` int NULL,
```
javierdlrm, line 2001: Why not `int(11)`?
Can you replace `NULL` with `DEFAULT NULL`? it's the same, just more readable :)

Same in the other files.

files/default/sql/ddl/3.2.0__initial_tables.sql
```diff
@@ -1995,21 +1995,29 @@ CREATE TABLE IF NOT EXISTS `job_alert` (
...
+                                                             `status` varchar(45) NOT NULL,
+                                                             `type` varchar(45) NOT NULL,
+                                                             `severity` varchar(45) NOT NULL,
+                                                             `receiver` int NOT NULL,
+                                                             `created` timestamp DEFAULT CURRENT_TIMESTAMP NOT NULL,
+                                                             `feature_view_id` int NULL,
```
javierdlrm, line 2007: Same. `int(11) DEFAULT NULL` ?

files/default/sql/ddl/3.2.0__initial_tables.sql
```diff
@@ -1995,21 +1995,29 @@ CREATE TABLE IF NOT EXISTS `job_alert` (
...
-                                                     CONSTRAINT `fk_feature_group_alert_1` FOREIGN KEY (`receiver`) REFERENCES `alert_receiver` (`id`) ON DELETE CASCADE,
-                                                     CONSTRAINT `fk_feature_group_alert_2` FOREIGN KEY (`feature_group_id`) REFERENCES `feature_group` (`id`) ON DELETE CASCADE
-) ENGINE=ndbcluster DEFAULT CHARSET=latin1 COLLATE=latin1_general_cs;
+CREATE TABLE IF NOT EXISTS `hopsworks`.`feature_group_alert` (
+                                                             `id` int AUTO_INCREMENT
+                                                                 PRIMARY KEY,
+                                                             `feature_group_id` int NULL,
+                                                             `status` varchar(45) NOT NULL,
+                                                             `type` varchar(45) NOT NULL,
+                                                             `severity` varchar(45) NOT NULL,
```
javierdlrm, line 2004: `COLLATE latin1_general_cs` ?

files/default/sql/ddl/3.2.0__initial_tables.sql
```diff
@@ -1995,21 +1995,29 @@ CREATE TABLE IF NOT EXISTS `job_alert` (
...
+                                                             `receiver` int NOT NULL,
```
javierdlrm, line 2005: Same. `int(11)`?

files/default/sql/ddl/updates/undo/3.2.0__undo.sql
```diff
@@ -66,3 +66,16 @@ DROP TABLE IF EXISTS `hopsworks`.`monitoring_window_config`;
...
 DROP TABLE IF EXISTS `hopsworks`.`job_schedule`;
+
+-- alert feature monitoring changes
+ALTER TABLE hopsworks.feature_group_alert
+    DROP FOREIGN KEY `fk_feature_group_alert_fv`,
+    DROP CONSTRAINT `unique_feature_view_status`,
+    DROP COLUMN `entity_type`,
+    DROP COLUMN `feature_view_id`,
+    DROP FOREIGN KEY `fk_feature_group_alert_2`,
+    MODIFY COLUMN `feature_group_id` int not null;
```
javierdlrm, line 77: `int(11) NOT NULL`?

files/default/sql/ddl/3.2.0__initial_tables.sql
```diff
@@ -1995,21 +1995,29 @@ CREATE TABLE IF NOT EXISTS `job_alert` (
...
+                                                             `created` timestamp DEFAULT CURRENT_TIMESTAMP NOT NULL,
+                                                             `feature_view_id` int NULL,
+                                                             `entity_type` int NOT NULL,
```
javierdlrm, line 2008: If entity_type can only have two values (i.e., 0 for feature group and 1 for feature view), then it's better to use `TINYINT(1) NOT NULL`

ErmiasG (changes requested):
This is not a proper table:
1. It is called feature_group_alert but contains feature_view_alert too
2. There is no way of insuring that the type is consistent with what is in the table. (I can insert type FEATURE_GROUP but have feature view id)
3.  Same with ValidationRuleAlertStatus. I can add FEATURE_MONITOR_SHIFT_UNDETECTED but have feature group id.
4. Sooner or later we are going to have orphaned rows in this table b/c the foreign keys are nullable.

DhananjayMukhedkar:
> This is not a proper table:
> 
>     1. It is called feature_group_alert but contains feature_view_alert too
> 
>     2. There is no way of insuring that the type is consistent with what is in the table. (I can insert type FEATURE_GROUP but have feature view id)
> 
>     3. Same with ValidationRuleAlertStatus. I can add FEATURE_MONITOR_SHIFT_UNDETECTED but have feature group id.
> 
>     4. Sooner or later we are going to have orphaned rows in this table b/c the foreign keys are nullable.

I now have refactored to name it a `feature_store_alert` table and  changed the backend to use the JPA SINGLE_TABLE_STRATEGY with different EJB entity classes for feature group and feature view. I think this addresses the issue of type safety while inserting record and stream lines the code. The idea is to use the same table and API as we share the other columns data and it will be easy to reuse it incase of adding more alert entities in future e.g. for feature monitoring. For foreign keys being nullable, I think since its a child table wtih delete cascade conditions, it will ideally not have orphaned rows.

files/default/sql/ddl/updates/undo/3.2.0__undo.sql
```diff
@@ -66,3 +66,21 @@ DROP TABLE IF EXISTS `hopsworks`.`monitoring_window_config`;
 DROP TABLE IF EXISTS `hopsworks`.`statistics_comparison_config`;
 
 DROP TABLE IF EXISTS `hopsworks`.`job_schedule`;
+-- alert feature monitoring changes
+DROP TABLE IF EXISTS `hopworks`.`feature_store_alert`;
```
vatj, line 70: I don't think you can do that, you would lose all the data in the table if you drop and create.
You can delete all entries which have feature_view_id not null and then modify the table by dropping columns

vatj (changes requested):

ErmiasG (changes requested):
This does not fix the consistency issue. And the orphaned rows issue. I do not see why you cant create a new table for this.
Entity has nothing to do with having rest api. And I do not think this is easier to use in any way.

ErmiasG (approved):

PR #914, opened 2023-04-02
[HWORKS-480] Remove inode foreign key from storage connectors

SirOibaf:


files/default/sql/ddl/updates/3.2.0.sql
```diff
@@ -77,3 +78,113 @@ SET SQL_SAFE_UPDATES = 1;
 
 ALTER TABLE `hopsworks`.`cached_feature_group` DROP COLUMN `online_enabled`;
 ALTER TABLE `hopsworks`.`stream_feature_group` DROP COLUMN `online_enabled`;
+
+SET max_sp_recursion_depth=10;
+
+DROP PROCEDURE `path_resolver`;
```
moritzmeister, line 84: Shouldn't it be drop procedure if exists?

moritzmeister (approved):

files/default/sql/ddl/updates/3.2.0.sql
kouzant, line 84: You need to add `IF EXISTS`

files/default/sql/ddl/updates/3.2.0.sql
```diff
@@ -77,3 +78,113 @@ SET SQL_SAFE_UPDATES = 1;
...
+DROP FUNCTION `path_resolver_fn`;
```
kouzant, line 85: Similarly, add `IF EXISTS`

kouzant (changes requested):

kouzant (approved):

PR #926, opened 2023-04-18
[HWORKS-512] Add variable for docker cgroup parent

gibchikafa:


templates/default/sql/dml/undo/3.3.0__undo.sql.erb
```diff
@@ -0,0 +1 @@
+DELETE FROM `hopsworks`.`variables` WHERE `id`='docker_cgroup_parent';
```
kouzant, line 1: Undo files will be removed, so you may skip this

kouzant (changes requested):

kouzant (approved):

PR #929, opened 2023-04-19
[FSTORE-771] ux python proposal

vatj:
Moved the tables to 3.3


files/default/sql/ddl/3.2.0__initial_tables.sql
```diff
@@ -2276,4 +2276,4 @@ CREATE TABLE IF NOT EXISTS `hdfs_command_execution` (
   KEY `fk_hdfs_file_command_2_idx` (`src_inode_partition_id`,`src_inode_pid`,`src_inode_name`),
   CONSTRAINT `fk_hdfs_file_command_1` FOREIGN KEY (`execution_id`) REFERENCES `executions` (`id`) ON DELETE CASCADE,
   CONSTRAINT `fk_hdfs_file_command_2` FOREIGN KEY (`src_inode_partition_id`,`src_inode_pid`,`src_inode_name`) REFERENCES `hops`.`hdfs_inodes` (`partition_id`, `parent_id`, `name`) ON DELETE CASCADE
-) ENGINE=ndbcluster DEFAULT CHARSET=latin1 COLLATE=latin1_general_cs;
\ No newline at end of file
+) ENGINE=ndbcluster DEFAULT CHARSET=latin1 COLLATE=latin1_general_cs;
```
SirOibaf, line 2279: @vatj - Can you make sure you don't modify the 3.2.0 files? The tool we use to do database migrations compute checksums of the files. Adding removing new lines changes the checksum and creates issues.

files/default/sql/ddl/3.3.0__initial_tables.sql
```diff
@@ -2282,10 +2282,10 @@ CREATE TABLE IF NOT EXISTS `hdfs_command_execution` (
 CREATE TABLE IF NOT EXISTS `monitoring_window_config` (
     `id` INT(11) NOT NULL AUTO_INCREMENT,
     `window_config_type` INT(11) NOT NULL,
-    `specific_id` INT(11),
+    `training_dataset_id` INT(11),
     `time_offset` VARCHAR(63),
     `window_length` VARCHAR(63),
-    `row_percentage` INT(11),
+    `row_percentage` FLOAT,
```
javierdlrm, line 2288: How the row percentage will be represented with float? Is it from 0 to 1 for percentage and >1 number of rows?

files/default/sql/ddl/3.3.0__initial_tables.sql
```diff
@@ -2295,21 +2295,32 @@ CREATE TABLE IF NOT EXISTS `statistics_comparison_config` (
...
     `id` INT(11) NOT NULL AUTO_INCREMENT,
     `feature_group_id` INT(11),
     `feature_view_id` INT(11),
-    `feature_name` VARCHAR(63) COLLATE latin1_general_cs NOT NULL,
+    `feature_name` VARCHAR(63) COLLATE latin1_general_cs DEFAULT NULL,
     `description` VARCHAR(2000) COLLATE latin1_general_cs DEFAULT NULL,
     `name` VARCHAR(255) COLLATE latin1_general_cs NOT NULL,
     `enabled` BOOLEAN DEFAULT TRUE,
     `feature_monitoring_type` tinyint(4) NOT NULL,
     `alert_config` VARCHAR(63) COLLATE latin1_general_cs, -- dummy this should become ref to another table
```
DhananjayMukhedkar, line 2322: now that we are not persisting alerts in this table for V1, should we keep this field for future use anyway or remove it ?

files/default/sql/ddl/3.2.0__initial_tables.sql
vatj: I'll fix it thanks for pointing it out

files/default/sql/ddl/3.3.0__initial_tables.sql
vatj: from 0 to 1, it is made clear in the python client

PR #948, opened 2023-05-11
[HWORKS-529] Python version variable must be user readable

robzor92:


files/default/sql/ddl/updates/3.3.0.sql
```diff
@@ -100,3 +100,8 @@ ALTER TABLE `hopsworks`.`transformation_function` DROP KEY `inode_fn_fk`;
 
 ALTER TABLE `hopsworks`.`transformation_function` DROP COLUMN `inode_pid`, DROP COLUMN `partition_id`, DROP COLUMN `inode_name`;
 
+-- HWORKS-529: Python version variable must be user readable
+SET SQL_SAFE_UPDATES = 0;
```
SirOibaf, line 104: You don't have to do this, you can just change it in the DML files. DML files are templated and executed every time. So you should find where it's executed and just add the visibility there.

SirOibaf (changes requested):

files/default/sql/ddl/updates/3.3.0.sql
robzor92: Ah that's right! I changed the PR

SirOibaf (approved):

PR #953, opened 2023-05-19
[FSTORE-848] Update feature monitoring result table

vatj:


files/default/sql/ddl/updates/3.3.0.sql
```diff
@@ -305,9 +305,13 @@ CREATE TABLE IF NOT EXISTS `feature_monitoring_result` (
     `execution_id` INT(11) NOT NULL,
     `monitoring_time` timestamp DEFAULT CURRENT_TIMESTAMP,
     `shift_detected` BOOLEAN DEFAULT FALSE,
-    `detection_stats_id` INT(11) NOT NULL,
+    `detection_stats_id` INT(11),
     `reference_stats_id` INT(11),
-    `difference` FLOAT,
+    `difference` FLOAT DEFAULT NULL,
+    `specific_value` FLOAT DEFAULT NULL,
```
javierdlrm, line 311: Why do we need specific_value in this table, if we already have it in `monitoring_window_config` ?

vatj: It costs nothing to have it there and allows us some flexibility to enable editing the value over time

PR #955, opened 2023-05-22
[HWORKS-225] Hopsworks ha

ErmiasG:


providers/configure_server.rb
```diff
@@ -0,0 +1,256 @@
...
+    realm_name "#{realmname}"
+    jaas_context "jdbcRealm"
+    properties props
+    domain_name domain_name
+    password_file password_file
+    username username
+    admin_port admin_port
+    secure false
+    classname "com.sun.enterprise.security.auth.realm.jdbc.JDBCRealm"
+    not_if "#{asadmin_cmd} list-auth-realms #{target} | grep kthfsrealm"
```
kouzant, line 226: Replace the hardcoded realm with the variable

recipes/das_node.rb
```diff
@@ -0,0 +1,485 @@
+#rows_path = "#{domains_dir}/post.sql"
+
+case node['platform']
+when "rhel"
+  package "openssh-clients"
```
kouzant, line 5: Do we still need this?

recipes/das_node.rb
```diff
@@ -0,0 +1,485 @@
...
+end
+
+package "openssh-server"
```
kouzant, line 8: And this?

recipes/das_node.rb
```diff
@@ -0,0 +1,485 @@
...
+public_ip=my_public_ip()
```
kouzant, line 9: Use `my_private_ip` instead

recipes/das_node.rb
```diff
@@ -0,0 +1,485 @@
...
+admin_port = node['hopsworks']['admin']['port']
+username=node['hopsworks']['admin']['user']
+password=node['hopsworks']['admin']['password']
+
+config_nodes= node['hopsworks'].attribute?('config_node')? private_recipe_ips('hopsworks', 'config_node') : []
+
+current_version = node['hopsworks']['current_version']
+
+asadmin = "#{node['glassfish']['base_dir']}/versions/current/bin/asadmin"
+password_file = "#{domains_dir}/#{domain_name}_admin_passwd"
```
kouzant, line 26: If it's not too bothersome can you create helper functions to construct them? I've seen them in many places

recipes/das_node.rb
```diff
@@ -0,0 +1,485 @@
...
+  admin_port admin_port
+  target payara_config
+  asadmin asadmin
+  internal_port node['hopsworks']['internal']['port']
+  network_name "https-internal"
+  network_listener_name "https-int-list"
+  action :glassfish_configure_network
+end
+
+if node['hopsworks']['ha']['loadbalancer'].to_s == "true"
```
kouzant, line 115: Use `casecmp?` for comparison which is case-insensitive

recipes/das_node.rb
```diff
@@ -0,0 +1,485 @@
...
+  mode '0750'
+end
+
+directory node['hopsworks']['data_volume']['node_logs'] do
+  owner node['glassfish']['user']
+  group node['glassfish']['group']
+  mode '0750'
+end
+
+bash 'Move glassfish logs to data volume' do
```
kouzant, line 205: This is not needed I think. The purpose of this was to move old directories to the data volume. If you already create the directory in the data volume and symlink it then this snippet is not needed.

recipes/config_node.rb
```diff
@@ -0,0 +1,77 @@
...
+asadmin = "#{node['glassfish']['base_dir']}/versions/current/bin/asadmin"
+password_file = "#{domains_dir}/#{domain_name}_admin_passwd"
+username=node['hopsworks']['admin']['user']
+nodedir=node['glassfish']['nodes_dir']
+payara_config = "hopsworks-config"
+deployment_group = "hopsworks-dg"
+
+admin_port = node['hopsworks']['admin']['port']
+das_ip=private_recipe_ip('hopsworks', 'das_node')
+public_ip=my_public_ip()
```
kouzant, line 12: Replace with `my_private_ip`

providers/worker.rb
```diff
@@ -0,0 +1,140 @@
...
+              restart_domain_command: "#{restart_instance_command}",
+              stop_domain_command: "#{stop_instance_command}",
+              start_domain_timeout: systemd_start_timeout,
+              stop_domain_timeout: systemd_stop_timeout)
+    notifies :start, "service[#{service_name}]", :delayed
+  end
+
+  service service_name do
+    supports start: true, restart: true, stop: true, status: true
+    action [:enable]
```
kouzant, line 44: Add a guard here if `node['services']['enabled']` is "true"

kouzant (changes requested):

recipes/das_node.rb
ErmiasG: I thought about it too but I do not think it will be that useful.

kouzant (changes requested):

recipes/install.rb
```diff
@@ -465,7 +525,7 @@
 directory node['hopsworks']['domain1']['logs'] do
   recursive true
   action :delete
-  not_if { conda_helpers.is_upgrade }
+  only_if { Dir.entries(node['hopsworks']['domain1']['logs']).select {|entry| !(entry =='.' || entry == '..' || entry == '.gitkeep') }.empty? || !conda_helpers.is_upgrade }
```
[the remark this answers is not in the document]
ErmiasG: The dir is not empty.

kouzant (approved):

PR #957, opened 2023-06-01
[FSTORE-868] Support feature view statistics with event and commit times

javierdlrm:


vatj (approved):
We could add the change from use_event_time to is_event_time for FM config in this PR

PR #961, opened 2023-06-07
[HWORKS-555] Change the notary and checkpointing directory config for jupyter

gibchikafa:


jimdowling:
I don't think we should share notebook checkpoints with other users in the /tmp directory.
Is there an alternative direction in the secret directory that is not mounted?
Do you know why hopsfs mount is failing the writes?

gibchikafa:
The /tmp directory is the one in the user container

jimdowling:
You're right. it's local to the container, then it's ok.

jimdowling (approved):

PR #968, opened 2023-06-12
[APPEND][FSTORE-709] add stream fg key to cached_feature_extra_constraints

kennethmhc:


files/default/sql/ddl/updates/3.3.0.sql
```diff
@@ -145,6 +145,10 @@ ALTER TABLE `hopsworks`.`hdfs_command_execution` ADD UNIQUE KEY `uq_src_path` (`
 ALTER TABLE `hopsworks`.`feature_store_s3_connector` ADD COLUMN `arguments` VARCHAR(2000) DEFAULT NULL;
 
 --FSTORE-709
+DELETE FROM `cached_feature_extra_constraints` WHERE `stream_feature_group_id` NOT IN (
+    SELECT `id`
+    FROM `stream_feature_group`
+);
```
SirOibaf, line 151: I have a feeling that this is not going to work on app. I think there will be too many constraints left pending.

kennethmhc: @SirOibaf Do you mean 
```
    SELECT `id`
    FROM `stream_feature_group`
```
has too many record, and it won't work?

SirOibaf: Yeah, one delete transaction can delete up to 32K records. I don't know how many contraints are pending on app.

SirOibaf:
This PR is superseeded by this one: https://github.com/logicalclocks/hopsworks-chef/pull/980

PR #973, opened 2023-06-19
[HWORKS-607] Remove inode foreign key from feature store statistic

javierdlrm:


files/default/sql/ddl/updates/3.3.0.sql
```diff
@@ -163,4 +163,13 @@ ALTER TABLE `hopsworks`.`feature_view`
 ALTER TABLE `cached_feature_extra_constraints` ADD KEY stream_feature_group_fk (`stream_feature_group_id`);
 ALTER TABLE `cached_feature_extra_constraints` ADD CONSTRAINT `stream_feature_group_fk1` FOREIGN KEY (`stream_feature_group_id`) REFERENCES `stream_feature_group` (`id`) ON DELETE CASCADE ON UPDATE NO ACTION;
 
+-- HWORKS-607: Remove inode from statistics
+ALTER TABLE `hopsworks`.`feature_store_statistic`
+  DROP FOREIGN KEY `inode_fk`,
+  DROP COLUMN `inode_pid`,
+  DROP COLUMN `inode_name`,
+  DROP COLUMN `partition_id`,
+  ADD COLUMN `file_path` VARCHAR(1000) COLLATE latin1_general_cs NOT NULL;
```
SirOibaf, line 172: This migration is wronge. You need to set the value of the `file_path` variable based on the existing inode before dropping it.

Also it seems here that you are not dropping the index, so I don't think you'll be able to drop the columns. See previous examples of migrations to see what you need to do.

SirOibaf (changes requested):

files/default/sql/ddl/updates/3.3.0.sql
javierdlrm: Fixed

SirOibaf (approved):

PR #974, opened 2023-06-22
[FSTORE-921] Resolve serving key in feature view

kennethmhc:


davitbzh (changes requested):
I think `undo` is missing

davitbzh (approved):

PR #976, opened 2023-06-28
[FSTORE-840] Use existing Kafka cluster

bubriks:
necessary as otherwise user wouldn't be able to update connector

### DIFF
--- a/templates/default/sql/dml/3.3.0.sql.erb
+++ b/templates/default/sql/dml/3.3.0.sql.erb
@@ -7,6 +7,3 @@ REPLACE INTO `hopsworks`.`variables`(`id`, `value`) VALUES ("hopsfsmount_apparmo
 REPLACE INTO `hopsworks`.`variables`(`id`, `value`) VALUES ("apply_hopsfsmount_apparmor_profile_kube", "<%= @apparmor_enabled %>");
 
 REPLACE INTO `hopsworks`.`variables`(`id`, `value`, `visibility`) VALUES ("ha_enabled", "<%= @ha_enabled %>", 1);
-
--- To ensure that enable_kafka_storage_connectors is updated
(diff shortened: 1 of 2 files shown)


templates/default/sql/dml/3.3.0.sql.erb
```diff
@@ -6,4 +6,7 @@ REPLACE INTO `hopsworks`.`variables`(`id`, `value`, `visibility`) VALUES ("jupyt
 REPLACE INTO `hopsworks`.`variables`(`id`, `value`) VALUES ("hopsfsmount_apparmor_profile", "<%= node['hops']['docker']['hopsfsmount-apparmor-profile'] %>");
 REPLACE INTO `hopsworks`.`variables`(`id`, `value`) VALUES ("apply_hopsfsmount_apparmor_profile_kube", "<%= @apparmor_enabled %>");
 
-REPLACE INTO `hopsworks`.`variables`(`id`, `value`, `visibility`) VALUES ("ha_enabled", "<%= @ha_enabled %>", 1);
\ No newline at end of file
+REPLACE INTO `hopsworks`.`variables`(`id`, `value`, `visibility`) VALUES ("ha_enabled", "<%= @ha_enabled %>", 1);
+
+-- To ensure that enable_kafka_storage_connectors is updated
+REPLACE INTO `hopsworks`.`variables`(`id`, `value`, `visibility`) VALUES ("enable_kafka_storage_connectors", "<%= node['hopsworks']['enable_kafka_storage_connectors'] %>", 1);
```
SirOibaf, line 12: You actually don't have to do this. All the DML files are executed at every update. So you just have to change the attributes/defaults.rb file and the value is going to be updated automatically.

SirOibaf (changes requested):

SirOibaf (approved):

PR #989, opened 2023-07-13
[FSTORE-946] Fix orphan statistics when deleting FGs

javierdlrm:


javierdlrm:
I have also deleted the file 3.4.0.sql.

SirOibaf (approved):

PR #991, opened 2023-07-13
[FSTORE-857] Overwrite hsfs_utils in jobs when upgrading to 3.3.0

maismail:
Fixes https://hopsworks.atlassian.net/browse/FSTORE-857

### DIFF
--- a/files/default/sql/ddl/updates/3.3.0.sql
+++ b/files/default/sql/ddl/updates/3.3.0.sql
@@ -243,3 +243,9 @@ ALTER TABLE `hopsworks`.`git_executions` ADD COLUMN `hostname` VARCHAR(128) COLL
 ALTER TABLE `hopsworks`.`feature_store_statistic`
   DROP FOREIGN KEY `fg_ci_fk_fss`,
   DROP KEY `feature_group_commit_id_fk`;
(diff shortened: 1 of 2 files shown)


files/default/sql/ddl/updates/3.3.0.sql
```diff
@@ -238,3 +238,7 @@ ALTER TABLE `hopsworks`.`feature_store_statistic`
 
 -- Git executions hostname
 ALTER TABLE `hopsworks`.`git_executions` ADD COLUMN `hostname` VARCHAR(128) COLLATE latin1_general_cs NOT NULL DEFAULT "localhost";
+
+-- FSTORE-857
+UPDATE hopsworks.jobs SET json_config=JSON_SET(json_config, '$.appPath', 'hdfs:///user/spark/hsfs-utils-3.3.0-RC0.jar') WHERE JSON_EXTRACT(json_config, '$.appPath') like '"hdfs:///user/spark/hsfs-utils-%.jar"';
```
SirOibaf, line 243: I think you need to disable safe updates here because you don't have a where condition on the PK

SirOibaf (approved):

PR #1015, opened 2023-07-24
[FSTORE-952] Single Kafka topic per project

bubriks:


files/default/sql/ddl/updates/3.4.0.sql
```diff
@@ -93,3 +93,14 @@ ALTER TABLE `hopsworks`.`stream_feature_group` ADD COLUMN `timetravel_format` IN
...
+-- FSTORE-952: Single Kafka topic per project
+ALTER TABLE `hopsworks`.`project_topics` MODIFY COLUMN `subject_id` int(11) DEFAULT NULL;
+
+UPDATE `hopsworks`.`subjects`
+SET `subject` = REGEXP_REPLACE(`subject`, "(.*)_(.*)_(.*)_(.*)_onlinefs",'$3_$4')
+WHERE REGEXP_SUBSTR(`subject`, "(.*)_(.*)_(.*)_(.*)_onlinefs");
+
+ALTER TABLE `hopsworks`.`feature_group` ADD COLUMN `use_project_topic` BOOLEAN DEFAULT TRUE;
+
+UPDATE `hopsworks`.`feature_group` SET `use_project_topic` = FALSE;
```
SirOibaf, line 106: Need to disable safe updates.

files/default/sql/ddl/updates/3.4.0.sql
```diff
@@ -93,3 +93,14 @@ ALTER TABLE `hopsworks`.`stream_feature_group` ADD COLUMN `timetravel_format` IN
 
 -- HWORKS-705: Chef should not pre-register the hosts
 ALTER TABLE `hopsworks`.`hosts` MODIFY COLUMN `num_gpus` tinyint(1) DEFAULT '0';
+
+-- FSTORE-952: Single Kafka topic per project
+ALTER TABLE `hopsworks`.`project_topics` MODIFY COLUMN `subject_id` int(11) DEFAULT NULL;
+
+UPDATE `hopsworks`.`subjects`
+SET `subject` = REGEXP_REPLACE(`subject`, "(.*)_(.*)_(.*)_(.*)_onlinefs",'$3_$4')
+WHERE REGEXP_SUBSTR(`subject`, "(.*)_(.*)_(.*)_(.*)_onlinefs");
```
SirOibaf, line 102: I'm not sure I'm following what's going on here but in general I have 3 questions:
- Why is this change needed?
- Shouldn't this be applied also to the kafka topic used for stream feature group not online?
- You need to disable the safe update here as well as you don't have a condition on the primary key.

SirOibaf (changes requested):

files/default/sql/ddl/updates/3.4.0.sql
bubriks: 1. This change was done to simplify/standardize naming of subject, now for feature group schemas it is the entity name (aka. "fgName_fgVersion") instead of topic name (aka. "projectId_fgId_fgName_fgVersion"). This is relatively insignificant change and could be undone if you want.
2. yes thats true will fix.
3. ok thanks

PR #1019, opened 2023-08-04
[HWORKS-679] Delete project docker images from registry with tags

gibchikafa:


o-alex (approved):

o-alex:
@robzor92 are there more things to clean from docker that you can think of?

o-alex:
any reasons this is not merged?

PR #1020, opened 2023-08-04
[HWORKS-684] Separate job scheduler from feature monitoring PR

SirOibaf:


files/default/sql/ddl/3.4.0__initial_tables.sql
```diff
@@ -2257,3 +2257,16 @@ CREATE TABLE `serving_key` (
                                CONSTRAINT `feature_view_serving_key_fk` FOREIGN KEY (`feature_view_id`) REFERENCES `feature_view` (`id`) ON DELETE CASCADE ON UPDATE NO ACTION,
                                CONSTRAINT `feature_group_serving_key_fk` FOREIGN KEY (`feature_group_id`) REFERENCES `feature_group` (`id`) ON DELETE CASCADE ON UPDATE NO ACTION
 ) ENGINE=ndbcluster DEFAULT CHARSET=latin1 COLLATE=latin1_general_cs;
+
+CREATE TABLE IF NOT EXISTS `job_schedule` (
+    `id` int NOT NULL AUTO_INCREMENT,
+    `job_id` int NOT NULL,
+    `start_date_time` timestamp NOT NULL,
+    `end_date_time` timestamp,
```
robzor92, line 2265: Looking at this table defintion `end_date_time` allows null, but entity field mapping says @Basic(optional = false)

robzor92 (changes requested):

files/default/sql/ddl/3.4.0__initial_tables.sql
SirOibaf: `end_date_time` can be null. I fixed the field mapping.

PR #1023, opened 2023-08-16
[HWORKS-135] Models backend should store metadata in tables instead of opensearch

robzor92:


files/default/sql/ddl/3.4.0__initial_tables.sql
```diff
@@ -2257,3 +2257,29 @@ CREATE TABLE `serving_key` (
...
+  `user_full_name` VARCHAR(255) DEFAULT NULL,
+  `created` timestamp DEFAULT NULL,
+  `description` VARCHAR(255) DEFAULT NULL,
+  `metrics` VARCHAR(1000) DEFAULT NULL,
+  `program` VARCHAR(1000) DEFAULT NULL,
+  `framework` VARCHAR(1000) DEFAULT NULL,
+  `environment` VARCHAR(1000) DEFAULT NULL,
+  `experiment_id` VARCHAR(1000) DEFAULT NULL,
+  `experiment_project_name` VARCHAR(100) DEFAULT NULL,
+  `model_registry_id` int(11) NOT NULL,
```
o-alex, line 2282: shouldn't the model registry id be part of the model?
also maybe the user full name should be a user id foreign key?

robzor92: I'm not sure I will keep the model_registry_id, because it's pretty much just the project_id so would not make sense to store it I think.

robzor92: On the user full name I just dumped what we had in elastic there for now. I was thinking it would be better to store the user FK there instead as well but started thinking of how to migrate it first. Will look more at it later on when closer to merging.

robzor92: I changed the user_full_name to be a user_id foreign key and added the migration in expat. Also the model_registry_id is removed and is now simply the project_id and not stored in the database.

files/default/sql/ddl/3.5.0__initial_tables.sql
```diff
@@ -2356,3 +2356,29 @@ CREATE TABLE `feature_store_keyword` (
...
+  PRIMARY KEY (`id`),
+  CONSTRAINT `model_project_fk` FOREIGN KEY (`project_id`) REFERENCES `project` (`id`) ON DELETE CASCADE ON UPDATE NO ACTION
+) ENGINE=ndbcluster DEFAULT CHARSET=latin1 COLLATE=latin1_general_cs;
+
+CREATE TABLE IF NOT EXISTS `model_version` (
+  `model_id` int(11) NOT NULL,
+  `version` int(11) NOT NULL,
+  `user_id` int(10) NOT NULL,
+  `created` timestamp DEFAULT NULL,
+  `description` VARCHAR(1000) DEFAULT NULL,
```
o-alex, line 2374: sync with hopsworks-ee DAO

robzor92: should be synced now

o-alex (approved):

PR #1026, opened 2023-08-26
[HWORKS-704][HWORKS-705][HWORKS-354]

SirOibaf:
…doesn't exists

### DIFF
--- a/templates/default/sql/dml/0.1.0.sql.erb
+++ b/templates/default/sql/dml/0.1.0.sql.erb
@@ -83,13 +83,7 @@ REPLACE INTO variables(`id`, `value`) VALUES ("epipe_ip", "<%= @epipe_ip %>");
 
 REPLACE INTO variables(`id`, `value`) VALUES ("livy_ip", "<%= @livy_ip %>");
 
-REPLACE INTO variables(`id`, `value`) VALUES ("yarn_ui_ip", "<%= @yarn_ui_ip %>");
-
 REPLACE INTO variables(`id`, `value`) VALUES ("yarn_ui_port", "<%= node['hops']['rm']['http_port'] %>");
-
-REPLACE INTO variables(`id`, `value`) VALUES ("hdfs_ui_ip", "<%= @hdfs_ui_ip %>");
-
-REPLACE INTO variables(`id`, `value`) VALUES ("hdfs_ui_port", "<%= @hdfs_ui_port %>");
 
 REPLACE INTO variables(`id`, `value`) VALUES ("file_preview_image_size", "<%= node['hopsworks']['file_preview_image_size'] %>");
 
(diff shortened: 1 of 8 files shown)


templates/default/sql/dml/0.1.0.sql.erb
```diff
@@ -53,11 +53,6 @@ REPLACE INTO `user_group` (`uid`, `gid`) VALUES (10001, 1007);
 REPLACE INTO variables(`id`, `value`) VALUES ("twofactor_auth", "<%= node['hopsworks']['twofactor_auth'] %>");
 REPLACE INTO variables(`id`, `value`) VALUES ("twofactor-excluded-groups", "<%= node['hopsworks']['twofactor_exclude_groups'] %>");
 
--- kmon
-
-INSERT INTO hosts(hostname, host_ip) VALUES <%= @hosts %> ON DUPLICATE KEY UPDATE cores=NULL;
```
kouzant, line 58: We still need the hosts table, don't we?

kouzant (changes requested):

templates/default/sql/dml/0.1.0.sql.erb
SirOibaf: I didn't remove the host table. I just removed the insert - I also changed Hopsworks to create the host if it doesn't exists: https://github.com/logicalclocks/hopsworks-ee/pull/1524

templates/default/sql/dml/0.1.0.sql.erb
```diff
@@ -72,30 +67,8 @@ REPLACE INTO variables(`id`, `value`) VALUES
...
 
-REPLACE INTO variables(`id`, `value`) VALUES ("rm_ip", "<%= @rm_ip %>");
-
-REPLACE INTO variables(`id`, `value`) VALUES ("rm_port", "<%= @rm_port %>");
-
-REPLACE INTO variables(`id`, `value`) VALUES ("epipe_ip", "<%= @epipe_ip %>");
-
-REPLACE INTO variables(`id`, `value`) VALUES ("livy_ip", "<%= @livy_ip %>");
-
-REPLACE INTO variables(`id`, `value`) VALUES ("yarn_ui_ip", "<%= @yarn_ui_ip %>");
```
maismail, line 91: not urgent, but we should also clean up this on in ec2-init as well 
https://github.com/logicalclocks/ec2-init/blob/master/ec2init/plans/preinit/pre_inits.py#L387

templates/default/sql/dml/0.1.0.sql.erb
maismail: that should be fine, we already delete the old hosts and re-register new head when initialising the cluster 
https://github.com/logicalclocks/ec2-init/blob/master/ec2init/plans/preinit/pre_inits.py#L592

maismail (approved):

PR #1028, opened 2023-08-27
[HWORKS-706] Switch opensearch-dashboard to use consul instead of IPs

SirOibaf:


templates/default/sql/dml/0.1.0.sql.erb
```diff
@@ -177,7 +177,6 @@ REPLACE INTO variables(`id`, `value`) VALUES ("max_cpu_request_size","<%= node['
 -- kibana, logstash,
 --
 
-REPLACE INTO variables(`id`, `value`) VALUES ("kibana_ip", "<%= @kibana_ip %>");
```
maismail, line 180: not urgent, but we should also clean up this on in ec2-init as well
https://github.com/logicalclocks/ec2-init/blob/master/ec2init/plans/preinit/pre_inits.py#L389

maismail (approved):

PR #1038, opened 2023-09-11
[FSTORE-862][APPEND] Upgrade fix

bubriks:


files/default/sql/ddl/updates/3.4.0.sql
```diff
@@ -158,6 +158,18 @@ ALTER TABLE `hopsworks`.`feature_group` ADD COLUMN `topic_name` VARCHAR(255) DEF
 ALTER TABLE `hopsworks`.`project` ADD COLUMN `topic_name` VARCHAR(255) DEFAULT NULL;
 
 SET SQL_SAFE_UPDATES = 0;
+-- deletes unused topics (fg was deleted)
```
maismail, line 161: why do we endup with alot of orphaned topics in the first place?

bubriks: It happens because we dont delete topics from online enabled feature groups (with suffix _onlinefs). We keep kafka topics in order to avoid consumers (onlinefs) getting blocked. This is something we should fix, but right now there is no solution.

files/default/sql/ddl/updates/3.4.0.sql
```diff
@@ -158,6 +158,23 @@ ALTER TABLE `hopsworks`.`feature_group` ADD COLUMN `topic_name` VARCHAR(255) DEF
...
+DELETE
+FROM `hopsworks`.`project_topics`
+WHERE REGEXP_SUBSTR(`topic_name`, "^([0-9]+)_([0-9]+)_(.+)_([0-9]+)(_onlinefs|$)")
+  AND REGEXP_REPLACE(`topic_name`, "^([0-9]+)_([0-9]+)_(.+)_([0-9]+)(_onlinefs|$)", '$2') NOT IN (SELECT `feature_group`.`id` FROM `feature_group`);
+
+-- remove the foreign key to subject from data ingestion topics (henceforth not needed)
+UPDATE `hopsworks`.`project_topics`
```
maismail, line 168: I'm not really sure why do we need to remove the fk in here

bubriks: technically not necessary to fix upgrade failing, but if not done it could cause problem when deleting online enable fg here: https://github.com/logicalclocks/hopsworks-ee/pull/1544/files (it would throw exception, since we would try to delete subject that is referenced by project_topics)

maismail (approved):

PR #1039, opened 2023-09-11
[HWORKS-734] Remove jdbc pool when initializing online feature store …

SirOibaf:
…database for project

### DIFF
--- a/recipes/default.rb
+++ b/recipes/default.rb
@@ -556,6 +556,16 @@ if exists_local("hops_airflow", "default")
   end
 end
 
+# Drop Existing featureStore connection pool and recreate it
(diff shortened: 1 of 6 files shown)


kennethmhc (approved):
LGTM!

recipes/default.rb
```diff
@@ -556,33 +556,6 @@
   end
 end
 
-# Drop Existing featureStore connection pool and recreate it
-glassfish_asadmin "delete-jdbc-connection-pool --cascade featureStorePool" do
```
kouzant, line 560: Shall we keep this to remove the connection pool on upgrades?

SirOibaf: Yes we should probably do that, otherwise I think the metrics are going to be messed up.

PR #1055, opened 2023-09-26
[HWORKS-762] Don't try to undeploy hopsworks-* current_version if it'…

SirOibaf:
…s not deployed

### DIFF
--- a/recipes/default.rb
+++ b/recipes/default.rb
@@ -374,6 +374,7 @@ if current_version.eql?("") == false
     keep_state true
     enabled true
     secure true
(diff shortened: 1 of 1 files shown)


recipes/default.rb
```diff
@@ -374,6 +374,7 @@
     keep_state true
     enabled true
     secure true
+    only_if "#{asadmin_cmd} list-applications --type ejb #{deployment_group} | grep -w \"hopsworks-ear:#{node['hopsworks']['current_version']}\""
```
kouzant, line 377: Considering the `target "domain"` above, is the `#{deployment_group}` correct?

kouzant (approved):

PR #1063, opened 2023-10-09
[HWORKS-784] Remove airflow jdbc metrics configuration

gibchikafa:


recipes/default.rb
```diff
@@ -507,40 +507,6 @@
...
-
-  # Drop Existing airflowPool connection pool and recreate it
-  glassfish_asadmin "delete-jdbc-connection-pool --cascade airflowPool" do
-    domain_name domain_name
-    password_file password_file
-    username username
-    admin_port admin_port
-    secure false
-    only_if "#{asadmin_cmd} list-jdbc-connection-pools | grep 'airflowPool$'"
-  end
```
SirOibaf, line 524: Leave this one so that it gets cleaned up during an upgrade.

SirOibaf (commented):
the install.rb recipe needs some cleanup as well.

SirOibaf (approved):

PR #1069, opened 2023-11-02
[HWORKS-826] Docker build should push to both region registries

SirOibaf:


kouzant (approved):
Needs fixing. I will fix it as soon as it gets merged

PR #1084, opened 2023-12-07
[HWORKS-860] hopsworks realm

ErmiasG:


providers/configure_server.rb
```diff
@@ -196,19 +196,13 @@
   admin_port=new_resource.admin_port
   target=new_resource.target
   asadmin=new_resource.asadmin
-  realmname = "kthfsrealm"
+  realmname = "hopsworksrealm"
```
kouzant, line 199: Why don't you use this variable everywhere instead of `hopsworksrealm`

recipes/install.rb
```diff
@@ -556,6 +556,17 @@
   action :create_if_missing
 end
 
+cauth = File.basename(node['hopsworks']['cauth_url'])
+
+remote_file "#{theDomain}/lib/#{cauth}" do
+  node['install']['enterprise']['install'].casecmp?("true") ? headers( "Authorization"=>"Basic #{ Base64.encode64("#{node['install']['enterprise']['username']}:#{node['install']['enterprise']['password']}").gsub("\n", "") }" ) : headers({})
```
kouzant, line 562: Use this function https://github.com/logicalclocks/hops-hadoop-chef/blob/master/recipes/install.rb#L326 Also, set the sensitivity flag

kouzant (changes requested):

kouzant (approved):

PR #1099, opened 2024-01-30
[FSTORE-612] Support for feature monitoring

javierdlrm:


files/default/sql/ddl/3.7.0__initial_tables.sql
```diff
@@ -407,6 +407,24 @@ CREATE TABLE IF NOT EXISTS `feature_group_statistics` (
 
 /*!40101 SET character_set_client = @saved_cs_client */;
 
+--
+-- Table structure for table `feature_group_statistics`
+--
+
+CREATE TABLE IF NOT EXISTS `feature_view_statistics` (
```
SirOibaf, line 414: why is it back?

files/default/sql/ddl/3.7.0__initial_tables.sql
```diff
@@ -451,6 +469,18 @@ CREATE TABLE IF NOT EXISTS `feature_group_descriptive_statistics` ( -- many-to-m
     CONSTRAINT `fgds_fds_fk` FOREIGN KEY (`feature_descriptive_statistics_id`) REFERENCES `feature_descriptive_statistics` (`id`) ON DELETE NO ACTION ON UPDATE NO ACTION
 ) ENGINE = ndbcluster DEFAULT CHARSET = latin1 COLLATE = latin1_general_cs;
 
+--
+-- Table structure for table `feature_view_descriptive_statistics`
+--
+
+CREATE TABLE IF NOT EXISTS `feature_view_descriptive_statistics` ( -- many-to-many relationship for legacy feature_view_statistics table
```
SirOibaf, line 476: Also this one

metadata.rb
```diff
@@ -1047,3 +1047,8 @@
 attribute "hopsworks/statistics/statistics_cleaner_interval_ms",
         :description => "How often the statistics cleaner is triggered.",
         :type => "string"
+
+# Feature monitoring
+attribute "hopsworks/enable_feature_monitoring",
+        :description => "Whether to enable feature monitoring with scheduling and alerting",
```
SirOibaf, line 1053: Add the default value in the description please.

SirOibaf (changes requested):

files/default/sql/ddl/3.7.0__initial_tables.sql
javierdlrm: There were two options, either expanding the training_dataset_statistics table (e.g., with window times columns, another unique key, etc) or having the feature_view_statistics table back. I tried the first option but it was getting too complex. Having the feature_view_statistics table back is definitely a better option.

Part of the complexity was also due to the fact that we need to store statistics computed on data obtained from a feature view, but that do not relate to any training dataset (no training dataset version).

files/default/sql/ddl/3.7.0__initial_tables.sql
```diff
@@ -2473,3 +2505,82 @@ CREATE TABLE IF NOT EXISTS `model_version` (
   CONSTRAINT `model_fk` FOREIGN KEY (`model_id`) REFERENCES `model` (`id`) ON DELETE CASCADE ON UPDATE NO ACTION
 ) ENGINE=ndbcluster DEFAULT CHARSET=latin1 COLLATE=latin1_general_cs;
 
+-- FSTORE-612: Feature monitoring
+CREATE TABLE IF NOT EXISTS `monitoring_window_config` (
+    `id` INT(11) NOT NULL AUTO_INCREMENT,
+    `window_config_type` INT(11) NOT NULL,
+    `training_dataset_version` INT(11),
+    `time_offset` VARCHAR(63),
+    `window_length` VARCHAR(63),
```
robzor92, line 2514: What was the motivation for storing a number in a varchar? Same question for time_offset.

robzor92 (changes requested):

attributes/default.rb
```diff
@@ -542,3 +542,9 @@
 ##
 default['hopsworks']['statistics']['statistics_cleaner_batch_size'] = "1000"
 default['hopsworks']['statistics']['statistics_cleaner_interval_ms'] = "900000"
+
+##
+## Feature Monitoring
+##
+default['hopsworks']['enable_feature_monitoring'] = "false"
```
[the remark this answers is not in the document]
javierdlrm: It was agreed a while ago as a preventive approach until feature monitoring is battle-tested.

files/default/sql/ddl/3.7.0__initial_tables.sql
javierdlrm: To allow for abbreviations such as "1w" (1 week), "2d" (2 days), etc.

javierdlrm: Possible values are validated with [these regex expressions](https://github.com/javierdlrm/hopsworks-ee/blob/25044d3635cbf0da16a0652d31349d72a9dc7c0c/hopsworks-common/src/main/java/io/hops/hopsworks/common/featurestore/FeaturestoreConstants.java#L171). There are unit tests covering it.

PR #1100, opened 2024-01-30
[HWORKS-977] Set hopsfsmount as default to jupyter_remote_fs_driver config

gibchikafa:


templates/default/sql/dml/3.7.0.sql.erb
```diff
@@ -13,4 +13,7 @@ DELETE FROM `hopsworks`.`variables` WHERE `id` = 'service_renew_token_3';
 DELETE FROM `hopsworks`.`variables` WHERE `id` = 'service_renew_token_4';
 
 -- HWORKS-302
-REPLACE INTO `hopsworks`.`variables`(`id`, `value`, `visibility`) VALUES ("enable_conda_install", "<%= node['hopsworks']['enable_conda_install'] %>", 1);
\ No newline at end of file
+REPLACE INTO `hopsworks`.`variables`(`id`, `value`, `visibility`) VALUES ("enable_conda_install", "<%= node['hopsworks']['enable_conda_install'] %>", 1);
+
+-- HWORKS-977
+REPLACE INTO `hopsworks`.`variables`(`id`, `value`, `visibility`) VALUES ("jupyter_remote_fs_driver", "<%= node['hopsworks']['jupyter']['remote_fs_driver'] %>", 1);
```
SirOibaf, line 19: You don't need to do this. All the DML files are executed every time, so this will already set the new default during upgrades: https://github.com/logicalclocks/hopsworks-chef/blob/8a18b69d9adbe328e3c1441dd7dc38183c2fa29a/templates/default/sql/dml/3.3.0.sql.erb#L5

SirOibaf (changes requested):

SirOibaf (changes requested):
You also need to change the value in the attributes/default.rb file

SirOibaf (approved):

PR #1120, opened 2024-03-06
[HWORKS-1059] Move API key generation in config_node if deploying HA configuration

kouzant:
https://hopsworks.atlassian.net/browse/HWORKS-1059

### DIFF
--- a/recipes/config_node.rb
+++ b/recipes/config_node.rb
@@ -81,7 +81,3 @@ hopsworks_worker "add_to_services" do
   action :add_to_services
   not_if "systemctl is-active --quiet #{service_name}"
 end
-
-hopsworks_configure_server "generate-int-api-key" do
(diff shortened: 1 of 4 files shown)


recipes/config_node.rb
```diff
@@ -80,4 +80,8 @@
   requires_authbind requires_authbind
   action :add_to_services
   not_if "systemctl is-active --quiet #{service_name}"
+end
+
+hopsworks_configure_server "generate-int-api-key" do
+  action :generate_internal_api_key
```
SirOibaf, line 86: @kouzant - Don't we need this in the das node as well?

kouzant: Actually `hopsworks::default` is sufficient both for HA and non-HA. In both configurations we deploy Hopsworks apps in default recipe

PR #1123, opened 2024-03-11
[HWORKS-1040] Add variable to configure timeout for conda commands

gibchikafa:
…ail the docker build


o-alex (approved):

attributes/default.rb
```diff
@@ -550,3 +550,5 @@
 ##
 default['hopsworks']['enable_feature_monitoring'] = "false"
 
+default['hopsworks']['library_install_timeout_minutes'] = 60
```
SirOibaf, line 553: Can you export this parameter in the metadata.rb file?

SirOibaf (changes requested):

PR #1133, opened 2024-04-16
[HWORKS-1130] Use base image to do conda search instead of using host's anaconda installation which has been removed

kouzant:
https://hopsworks.atlassian.net/browse/HWORKS-1130


templates/default/condasearch.sh.erb
```diff
@@ -1,15 +1,17 @@
...
 CHANNEL="${1//[^a-zA-Z0-9\.\-\_:\/\&~?]/}"
 if [ "$CHANNEL" != "defaults" ] ; then
    CHANNEL="$CHANNEL --override-channels"
 fi
+
 LIB="${2//[^a-zA-Z0-9\.\-\_]/}"
-<%= node['conda']['base_dir'] %>/bin/conda search -c $CHANNEL "${LIB}"  | grep -v Fetching | awk '{print $1","$2}' | grep -v "^\*"
+
+BASE_IMAGE_NAME="$3"
+docker run --rm $BASE_IMAGE_NAME conda search -c $CHANNEL "${LIB}" | grep -v Fetching | awk '{print $1","$2}' | grep -v "^\*"
```
robzor92, line 16: For other conda operations we mount the condarc: https://github.com/logicalclocks/conda-chef/blob/master/templates/default/condarc.erb

I.e when installing a library https://github.com/logicalclocks/hopsworks-ee/blob/5ff93f8336243c1ec1d858902c53ed182a85c045/hopsworks-common/src/main/java/io/hops/hopsworks/common/python/environment/DockerFileController.java#L170

I'm not sure if we also need to do it for search, maybe it was done automatically before when the script was running on the host where the config file was.

robzor92 (changes requested):

templates/default/condasearch.sh.erb
kouzant: I believe you're right. We need to use the condarc in under `glassfish` user because it has configuration for more Channels. There are some properties which are currently wrongly templated, I'll have a look.

templates/default/condaSearch.sh.erb
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+
+USER_HOME=${HW_USER_HOME:-/home/glassfish}
```
robzor92, line 3:
```suggestion
USER_HOME=${HW_USER_HOME:-/home/glassfish}
```

we should have a variable for the glassfish user name right?

robzor92 (changes requested):

robzor92 (approved):

PR #1140, opened 2024-05-14
[FSTORE-1285] Model Dependent Transformation Functions

manu-sj:
PR implements required tables for Model Dependent Transformations:

**Changes Done**

- Create new Table `feature_view_transformation_function` that maps feature view to corresponding transformation functions. It also stores the features that are used for the transformation as a string separated by commas in the column `features`.

- Modified table `training_dataset_feature` to remove column `transformation_function_id`. This is done because transformation functions are mapped using "feature view's" instead of "training data set features" to allow support for many to many transformations.

- Modified table `transformation_function` to include column `save_type`. `save_type` denotes how the transformation functions are saved. It can currently have two value - "feature_view" or "feature_group". Transformation function that are implicitly register in the feature view would have the `save_type` as "feature_view" and the transformation function saved into the feature store using the existing API would have the `save_type` as "feature_store"

### DIFF
--- a/files/default/sql/ddl/3.8.0__initial_tables.sql
+++ b/files/default/sql/ddl/3.8.0__initial_tables.sql
@@ -2080,7 +2080,7 @@ CREATE TABLE IF NOT EXISTS `transformation_function` (
                                                          `created` timestamp NULL DEFAULT CURRENT_TIMESTAMP,
                                                          `creator` int(11) NOT NULL,
                                                          `save_type`                         VARCHAR(255)    NOT NULL, 
-                                                         `statistics_argument_names`         VARCHAR(5000),
+                                                         `statistics_argument_names`         VARCHAR(5004),
                                                          PRIMARY KEY (`id`),
                                                          CONSTRAINT `feature_store_fn_fk` FOREIGN KEY (`feature_store_id`) REFERENCES `feature_store` (`id`) ON DELETE CASCADE ON UPDATE NO ACTION,
                                                          CONSTRAINT `creator_fn_fk` FOREIGN KEY (`creator`) REFERENCES `users` (`uid`) ON DELETE NO ACTION ON UPDATE NO ACTION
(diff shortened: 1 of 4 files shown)


files/default/sql/ddl/3.8.0__initial_tables.sql
```diff
@@ -2078,11 +2076,26 @@ CREATE TABLE IF NOT EXISTS `transformation_function` (
...
                                                          PRIMARY KEY (`id`),
                                                          CONSTRAINT `feature_store_fn_fk` FOREIGN KEY (`feature_store_id`) REFERENCES `feature_store` (`id`) ON DELETE CASCADE ON UPDATE NO ACTION,
                                                          CONSTRAINT `creator_fn_fk` FOREIGN KEY (`creator`) REFERENCES `users` (`uid`) ON DELETE NO ACTION ON UPDATE NO ACTION
 ) ENGINE = ndbcluster DEFAULT CHARSET = latin1 COLLATE = latin1_general_cs;
 
+--
+-- Table structure for table `feature_view_transformation_functions`
+--
+
+CREATE TABLE IF NOT EXISTS `feature_view_transformation_function` ( -- mapping transformation functions in feature view with required features and actual transformation function.
```
kennethmhc, line 2089: will you migrate the existing feature view which are using built-it transformation functions to this table?

manu-sj: Yes migration of existing feature views that using built-in transformations needs to be done.

kennethmhc: ok, will you add it here or in expat?

manu-sj: To migrate I believe I would have to :
1. Copy the modified transformation function source code into the transformation function folder of each project that has already been created.
2. Add entires into the table `feature_view_transformations` for each feature view that already has transformation functions.

So where do you think it would be best to do the migration. Can I do it here or should it be in expat.

kennethmhc: I think you need to delete the old transformation function and attach the new transformation function for each feature view which have transformation function. You can do it in expat.

manu-sj: Okey I will start working on that too.

kennethmhc (approved):

files/default/sql/ddl/3.8.0__initial_tables.sql
```diff
@@ -2077,6 +2077,7 @@ CREATE TABLE IF NOT EXISTS `transformation_function` (
                                                          `created` timestamp NULL DEFAULT CURRENT_TIMESTAMP,
                                                          `creator` int(11) NOT NULL,
                                                          `save_type`                         VARCHAR(255)    NOT NULL, 
+                                                         `statistics_argument_names`         VARCHAR(5000),
```
kennethmhc, line 2080: since there is "," joining the features, `5004` is better, then you can store 5 features of max length

kennethmhc (changes requested):

kennethmhc (approved):

PR #1144, opened 2024-05-31
[FSTORE-1420] Create tables for logging

bubriks:
https://hopsworks.atlassian.net/wiki/spaces/FST/pages/401670145/Resolving+partial+failures+in+Hopsworks+operations#Example-DB-diagram-(for-action-logging)


bubriks:
The option of using annotations/interceptors instead can be found here: https://github.com/logicalclocks/hopsworks-chef/pull/1147 (only one solution can be developed further)

